PR #19, opened 2017-10-04
Changed output format

sutterj:
Changed the output to print vertically instead of horizontally.


gkapfham:
Hello @sutterj. It looks like this pull request was merged into master. However, it commits Python cache files and directories into the repository. These files are derived and they should not be a part of the repository.

Issue #20, opened 2017-10-04
Repository Contains Derived Files


gkapfham:
The repository now contains a `gatorgrouper/__pycache__/` directory. If I understand correctly, this is a derived directory that does not need to be under version control. Should this directory be deleted?

sutterj:
My request was not yet merged into master.  The cache files were apart of merge #17 and #18.  They are also included in my commit, but it has not yet been merged.

delgrecoj:
This has been done (and added to a `.gitignore` file) in our branch for issue #13, as of this morning... for what that's worth.  Gotcha is that you have to remove the directory and commit that first before gitignoring it.

delgrecoj:
After some slight changes, the pull request received three approvals from project managers and was merged into master, so closing this issue.

PR #24, opened 2017-10-04
Removing cache files

sutterj:
I removed the pycache folder as per issue #20.  Look this over to make sure that I didn't mess anything up.


delgrecoj:
Should add a `.gitignore` file to prevent the files from coming back in the future as well.  Probably also worth ignoring the compiled `.pyc` files.
```shell
echo "__pycache__" >| .gitignore
echo "*.pyc" >| .gitignore
```

sutterj:
I think I got everything squared away.  Take a look and let me know.

PR #26, opened 2017-10-06
Feature/arg parser

milleriAC:
Added argument verification for the group size. Also added the ability to select the number of groups, but it is currently commented out. If there are no issues, we are ready to merge into task 2 branch.


castellanosr:
I think this looks good. If we want to implement the ability to select number of groups (the section we currently have commented out) we will need to add a little more code.

yeeunmariakim:
@milleriAC @castellanosr can we move the logic you put into `gatorgrouper.py` for checking the number of students into the `parse_gatorgrader_arguments.py` file?

sutterj:
There is an indentation error on line 28 of parse_gatorgrader_arguments.py.  Also, why is it called gatorgrader instead of gatorgrouper?

yeeunmariakim:
@sutterj that was a typo in the original files that I fixed in my own pull request. this branch is branching off of an early version of our feature branch. It will be fixed when I merge in my request.

castellanosr:
After moving the verification logic and attempting to make it work in the parse_gatorgrader_arguments.py file @shafferz and I found that it will not work there. It needs to stay where @milleriAC originally put it because of the order in which features run.

yeeunmariakim:
@castellanosr @shafferz with the new changes, the program does not run. Specifically, I get the error ```Traceback (most recent call last):
  File "gatorgrouper.py", line 28, in <module>
    if (list_size > 1 and gg_arguments.group_size < (list_size/2)):
NameError: name 'list_size' is not defined
MariaKim@Marias-MacBook-Pro ~/D/a/f/c/gator```

yeeunmariakim:
@shafferz @castellanosr debugging statements should be removed if they are not logged

yeeunmariakim:
@milleriAC @castellanosr @shafferz I moved the logic for the check to the `parse_gatorgrouper_arguments` file and merged in logging from the issue 4 branch.

Issue #23, opened 2017-10-04
Docstring Comments for Modules are Incorrect


gkapfham:
After the recent merge of the pull request, there are several new files in the repository. However, these files all have the same header comment at the top of the file. Each of these comments should be customized to the specific file and its purpose.

delgrecoj:
@ZachAndrews98 fixed as of #27.

PR #27, opened 2017-10-06
Fixed Docstring Comments

ZachAndrews98:
Moved the docstrings in the functions to the top of each file since that docstring is what the file's purpose is. Let me know if there is anything else I should add or formatting that should be changed.

### DIFF
--- a/defaults.py
+++ b/defaults.py
@@ -1,3 +1,4 @@
+""" Default Constants """
 DEFALT_STUDENT_FILE = "students.txt"
 GATORGROUPER_HOME = "GATORGROUPER_HOME"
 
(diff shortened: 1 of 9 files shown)


delgrecoj:
Looks fine to me, @ZachAndrews98.  Would close #23, although @gkapfham might open another for missing docstrings on functions -- we're really just getting ahead of ourselves here, trying to document things that haven't been implemented or haven't been patched just yet.  The function names effectively convey as much information as the docstrings did anyway, and relying on that until implementation is further along seems fine to me.

Will tag a couple other project managers to have them also review and approve this.

PR #29, opened 2017-10-11
Feature/import logging

yeeunmariakim:


yeeunmariakim:
@shafferz what do you think about this use of the logging framework?

shafferz:
Everything looks good to me! I also noted the changes we made today during our meetup, and they looked good as well.

PR #34, opened 2017-10-14
Feature/i 13 categorize students

sporrera:
Hi everyone,

So to get the categorization started, I first made a Google Form that asked students their preference of position (just as a way to test the code) which then exported into Google Sheets.  I have sent both of these to you via your allegheny email addresses.

In order to import the answers of the students into read_student_file.py, I had to download and save the Google Sheet as an excel file because I have no familiarity with Google API.  From there, I created a program within read_student_file.py that will print out everyone's last name followed by True and/or False boolean values corresponding to their answers, where True = Yes and False = No.

In a separate file when I practice running and implementing this code, it works fine.  However, I am having an issue returning the manipulated data in read_student_file.py.  My program will print out the appropriate data but will not return it as a list.  I included the error message that I am getting. 
error_message

When I run it independently of the gatorgrouper.py program, the output looks like this:

correct_output

I can continue to work on this tomorrow, I just wanted to update you and get any feedback that you have.


yeeunmariakim:
@sporrera if this feature is not ready to be merged into master, it may be better suited to be discussed in the issue tracker

delgrecoj:
This isn't quite working as intended, but I'm merging it into the feature branch, and then merging master into the feature branch... in the hopes that working with code from the other teams will be helpful here.

PR #36, opened 2017-10-14
Feature/i 14 categorize students

sporrera:
Pull request for read_student_file.py

This function returns a list of lists of the following structure:
[sporrer, False, True, False]
where the Trues and Falses are boolean values and not strings


delgrecoj:
this seems to be a duplicate pull request, into master instead of our feature branch.
I fixed the branch target on the other one, and it shows these commits, so I'm closing this one.

PR #39, opened 2017-10-15
Working group scoring file

ZachAndrews98:
Made slight changes to display.py and read_student_files.py in order to be able to use the basic structure I already had. The score_group function adds 1 to the score of the group each time there is a True value (which was the change in read_student_files). The score_groups function calls the score_group function for each group and checks if the lowest score is below a threshold (2/3 of the average score -1). If a group that meets this requirement exists the student with the highest value in the highest valued group and the student with the lowest value in the lowest valued group are switched. The change in display.py just makes it so that each student's characteristics aren't shown.


delgrecoj:
Might this switching / balancing functionality be better off disabled by default and exposed as an option via the command-line arguments?  Perhaps named `--rebalance` or similar?

My original thinking was that the scoring function would just return a score, without modifying anything, because the sudoku method might need to call it many times to evaluate potential groupings.

ZachAndrews98:
I think it is a good idea to add that rebalance argument so I will place that section of code into a separate file for if we decide to flesh that out and make it so that the functions just return the score(s).

PR #41, opened 2017-10-16
Issue 13 -- Grouping Methods (Random, Round-Robin)

delgrecoj:
Improved and working random and round-robin grouping methods.
Dropped sudoku method references in the interests of getting out of the way of the work other teams need to perform; also it's unlikely to be any better than the existing grouping methods.
This feature branch is working from the point of reading in the CSV data.


yeeunmariakim:
@sutterj have you modified the usage documentation to reflect the fact that Jesse has omitted the sudoku grouping method?

PR #46, opened 2017-10-16
Feature/i 6 usage documentation

sutterj:
I know that there are going to be more items that will need to be added as the program progresses, but this is a base with all of the command line arguments that I was given in addition to the initial setup.  Please look this over.  It might be better to merge it now, if there are no issues, and then create another branch to add things if needed.  This way other groups can begin to add things to a more updated version of the README.


sutterj:
@yeeunmariakim The instructions in the installation part were initially like that.  Do you still want me to change them?

yeeunmariakim:
@sutterj in the same way--with a `$` sign before the commands.

PR #58, opened 2017-10-17
Feature/i 3 test suites

bonnettd:
Finalized version of test suites in the test folder


delgrecoj:
Should `assignments.txt` be removed?  Also, the display function is already integrated into master in the `display.py` file isn't it?

bonnettd:
@delgrecoj Assignments.txt will be removed, as far as the cache file goes I would prefer that we just keep it. Also, I am unsure of what you are talking about with the display function? Does it need removed?

delgrecoj:
`display_student_groups` is already present in master, within the `display.py` file, so this version would be redundant... as for the cache file, generated files really shouldn't be comitted to master, but rather removed and gitignored

PR #59, opened 2017-10-17
Feature/i 3 test suites

bonnettd:
New pull request for issue 3 (Test suites). Assignments.txt has been removed, as well as display_student_groups and the cache


delgrecoj:
just for future reference -- pushing more commits to the branch a pull request is for will add them to that request and they'll show up on the pull request... no need to close and recreate pull requests in most cases

PR #63, opened 2017-10-17
Feature/fixing travis

Andre-Bryan:
Should work fixed the error. it had an error currently because the file was seeing the display_student_groups but that is not present in master so everything should work

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,6 +17,7 @@ install:
   - pip3 install pytest-flake8
   - pip3 install pytest-cov
   - pip3 install autopep8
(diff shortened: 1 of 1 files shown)


yeeunmariakim:
@Andre-Bryan can you pull from master to demonstrate that this fix will fix the Travis build in master?

Andre-Bryan:
passed

Andre-Bryan:
it passed
0 errors

Andre-Bryan:
need to try some other things because coveralls did not get anything like it was supposed to

PR #82, opened 2017-11-11
Changes to README.md

sporrera:
Upload changes to README.md that address both issues of instruction under Absent Students as well as Specify File Containing List of Students

These issues address:

- not specifying the exact filename of the .csv file (it could be either students.csv or student_list.csv, as preferred by the user)

- Removing commands that are supposed to remove absent students from the list of students.  The removed commands did not actually remove the names of the absent students from the list

### DIFF
--- a/README.md
+++ b/README.md
@@ -212,10 +212,10 @@ following command:
 pip3 install --user pytest-flake8
 ```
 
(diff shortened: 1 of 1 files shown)


sporrera:
Also updated the change to the instruction of how to run the test suite so that it can be more easily followed by the user

PR #84, opened 2017-11-13
I 72 num of groups

ZachAndrews98:
Use python3 gatorgrouper.py --num-group "number of groups" to run the program with a certain number of groups. Will not allow there to be more groups than students and allows there to be the same number of groups as students.


burrowss:
@ZachAndrews98 Can you check the test suites? There are 5 failing, but one of them is just flake8. Also the readme seems to fail mdl

PR #85, opened 2017-11-13
I 72 num of groups

ZachAndrews98:
Use python3 gatorgrouper.py --num-group "number of groups" to run the program with a certain number of groups. Will not allow there to be more groups than students and allows there to be the same number of groups as students. Works for both --round-robin method of creating groups as well as default random method.


millinj2:
@burrowss I said this previously in the issue but the flake8 test case is not passing for me which is giving us an X with Travis.

PR #87, opened 2017-11-14
changed the font of the group label for easier recognition

bonnettd:
The display() function now prints the group labels, for instance Group 1, in green, bold font, and is underlined. The font used for the students remains normal


bonnettd:
I am working to solve the errors with Travis CI

Issue #104, opened 2019-01-29
Adopt Better Standards for Project Code Linting


gkapfham:
The README file and the Travis CI configuration does not always adhere to best practices for code formatting. This project should use Black to format all of the source code. Additionally, it should ensure that the linting tools check this formatting so that a build in Travis CI will fail if the code is not correctly formatted.

ilikerustoo:
@aubreypc I have run black on all the files and this seems to cause pylint errors which I will look into. Otherwise I have manually fixed every pylint and flake8 issue in all the `.py` files in and outside the `tests` folder. I needed to do this to pass Travis when implementing linting tools and conversion to pipenv in the travis.yml file

ilikerustoo:
I have fixed more `pylint` issues after the restructure that are waiting to be merged from my branch `bugfix/ fixing-bugs-in-develop`. I have add `black --check` to Travis too.

Lancasterwu:
This issue has been resolved.

PR #122, opened 2019-02-02
Format .py files using black

aubreypc:
This PR implements some small changes to code formatting via `black ./*.py`.

This partially addresses issue #104 .

### DIFF
--- a/display.py
+++ b/display.py
@@ -8,12 +8,12 @@ def display_student_groups(student_groups):
     group_counter = 1
     place_counter = 0
     for student_group in student_groups:
-        #print("Group", group_counter)
-        print('\033[0;32m' + '\033[1m' + '\033[4m' + "Group", group_counter)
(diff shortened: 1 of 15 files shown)


gkapfham:
Hi @aubreypc, thanks for creating this PR. I think that it is essential for us to reformat the code. Before I complete a review of the PR, I'm wondering whether you think that it should also include the mechanism by which the code can be automatically reformatted and checked through Travis CI. I agree that we want to reformat the code, but we also don't want to have to keep opening this type of PR in the future. So, should this PR only be the reformat? Or, should it be the reformat and also all of the code needed to ensure that we do not need to worry about future reformats?

ilikerustoo:
@aubreypc I have run black on the .py files and am currently correcting `pylint` and `flake8` tests.

aubreypc:
Closing PR, as @ilikerustoo is handling this

Issue #115, opened 2019-01-29
Explain How to Install Python with Pyenv


gkapfham:
It is possible that people with have difficult installing and using GatorGrouper because they do not know how to install Python on their machine. While #81 suggests that we add a graphical user interface, and I agree that this is a good idea, we will have to focus first on finishing the tool and supporting the programmers who want to use it. This means that the README file should have documentation that explains how to setup a Python development environment, perhaps with Pyenv.

Lancasterwu:
This is finished in https://github.com/GatorEducator/gatorgrouper/pull/123 by the whole Team "Lancaster" with members @reibelj @thomad74 @wattob @ALEXANDERB82 @everitt-andrew @Lancasterwu . The pull request is https://github.com/GatorEducator/gatorgrouper/pull/123.

Michionlion:
We should keep in mind that this tool may be used by people in a Windows environment, and so `pyenv` or similar tools are not supported. Does your fix in #123 support this? If not, perhaps a separate issue can be opened to address that. In the meantime, since this issue is resolved, I'll close this.

Lancasterwu:
@Michionlion Hi, I think we suggested people to upgrade the python version by downloading from the official website. If they can not, they can try use `pyenv`. We also mentioned in different development environment the situation may be different such as the configuration files. I think it is probably the solution for that. What is your opinion?

Michionlion:
@Lancasterwu you could explicitly talk about the difference in Windows installation. However, if you feel like you've covered everything already, then that should be fine.

Issue #105, opened 2019-01-29
Test with Multiple Versions of Python Using Pyenv and Tox


gkapfham:
The tool is currently only being tested with an old version of Python. First, we should start developing and testing according to the features of a new version of Python. Next, we need to start performing test suite runs with multiple version of Python in Travis CI. I suggest that we investigate the combined used of Pyenv and Tox to support this type of testing. Overall, this type of testing will better ensure that our tool works for a wide variety of people who are using this tool

reibelj:
```reibelj@aldenv166:~/cs203/gatorgrouper$ pip install -r requirements.txt
Traceback (most recent call last):
  File "/usr/bin/pip", line 9, in <module>
    from pip import main
ImportError: cannot import name main```


When trying to run the program for the first time, this is the error I received from having a lower version of Python. To remedy this error, I upgraded Python and reinstalled pyenv.

**Comment by reibelj** [comment]
```
reibelj@aldenv166:~/cs203/gatorgrouper$ python3 gatorgrouper.py

GatorGrouper: Automatically Assign Students to Groups
https://github.com/GatorGrouper/gatorgrouper

Traceback (most recent call last):
  File "gatorgrouper.py", line 33, in <module>
    create_csv()
  File "/home/r/reibelj/cs203/gatorgrouper/spreadsheet.py", line 25, in create_csv
    sheet = client.open(DEFAULT_WORKBOOK).sheet1
  File "/home/r/reibelj/.local/lib/python3.7/site-packages/gspread/client.py", line 123, in open
    self.list_spreadsheet_files()
  File "/home/r/reibelj/.local/lib/python3.7/site-packages/gspread/client.py", line 96, in list_spreadsheet_files
    res = self.request('get', url, params=params).json()
  File "/home/r/reibelj/.local/lib/python3.7/site-packages/gspread/client.py", line 79, in request
    raise APIError(response)
gspread.exceptions.APIError: {
 "error": {
  "errors": [
   {
    "domain": "global",
    "reason": "insufficientPermissions",
    "message": "Insufficient Permission: Request had insufficient authentication scopes."
   }
  ],
  "code": 403,
  "message": "Insufficient Permission: Request had insufficient authentication scopes."
 }
}
```

There are several issues with the program that prevent it from running. This makes the issue of testing very difficult until these errors are resolved.
@Lancasterwu

ilikerustoo:
I can help with this. @Lancasterwu  @reibelj do you have anything so far?

[...]

PR #123, opened 2019-02-05
Added Installation of Pyenv in README.md

Lancasterwu:
Issue #115 
Added instruction on how to install Python by using Pyenv in README.md

This issue is handled by the whole Team "Lancaster" with members @reibelj @thomad74 @wattob @ALEXANDERB82 @everitt-andrew @Lancasterwu.

Please contact one of the team members if you have any questions.

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,7 +36,7 @@ Interested in learning more about the basics of virtual environments in Python
 [article](http://www.cs.allegheny.edu/sites/gkapfham/programming/research/idea/2017/07/14/Virtual-Environments/)
 to further develop your understanding of this topic.
 
-GatorGrouper requires user to use Python3. You can type `$ python --version` to check
+GatorGrouper requires user to use Python3. You can type `python --version` to check
 the current Python version on your devices. If you do not have the right Python version,
 or you can not download or upgrade Python on your devices, you can download
 [Pyenv](https://github.com/pyenv/pyenv) tool to set up the virtual environment of
@@ -46,7 +46,7 @@ To install Pyenv, you can use [Pyenv Installer](https://github.com/pyenv/pyenv-i
 By typing the command in terminal:
 
 ```shell
-$ curl https://pyenv.run | bash
+curl https://pyenv.run | bash
 ```
 
 You will successfully install Pyenv. Please make sure that you have the
@@ -63,7 +63,7 @@ Pyenv after restart your terminal window by typing `pyenv install 3.6.7`
 and `pyenv global 3.6.7` to install.
 
 After you finish this process, you should be able
-to type `$ python --version` and get `Python 3.6.7` in your terminal.
+to type `python --version` and get `Python 3.6.7` in your terminal.
 
 ## Initial Setup
 
--- a/README.md
+++ b/README.md
@@ -71,7 +71,7 @@ should restart your terminal window by typing `pyenv install 3.7.2`
 and `pyenv global 3.7.2`.
 
 After you finish this process, you should be able
-to type `python --version` and get 'Python 3.7.2` or other version after
+to type `python --version` and get `Python 3.7.2` or other version after
 `Python 3.6.8` in your terminal.
 
 ## Initial Setup
--- a/README.md
+++ b/README.md
@@ -36,14 +36,14 @@ Interested in learning more about the basics of virtual environments in Python
 [article](http://www.cs.allegheny.edu/sites/gkapfham/programming/research/idea/2017/07/14/Virtual-Environments/)
 to further develop your understanding of this topic.
 
-GatorGrouper requires users to use Python3. You can type `python --version` to check
+GatorGrouper requires users to use Python 3. You can type `python --version` to check
 the current Python version on your device. If you do not have the right Python version,
-or you can not download or upgrade Python on your devices, you can download
-[Pyenv](https://github.com/pyenv/pyenv) tool to set up the virtual environment of
+or you can not download or upgrade Python on your devices, you can download the
+[Pyenv](https://github.com/pyenv/pyenv) tool to set up a virtual environment for
 the newest Python version.
 
-To install Pyenv, you can use [Pyenv Installer](https://github.com/pyenv/pyenv-installer).
-By typing the command in terminal:
+To install Pyenv, you can use [Pyenv Installer](https://github.com/pyenv/pyenv-installer)
+by typing the command in terminal:
 
 ```shell
 curl https://pyenv.run | bash
@@ -58,8 +58,8 @@ eval "$(pyenv init -)"
 eval "$(pyenv virtualenv-init -)"
 ```
 
-Once Pyenv is installed, you can install Python 3.6.7 for your
-Pyenv after restart your terminal window by typing `pyenv install 3.6.7`
+Once Pyenv is installed, you can install the latest version of Python
+for your Pyenv after restart your terminal window by typing `pyenv install 3.6.7`
 and `pyenv global 3.6.7` to install.
 
 After you finish this process, you should be able
--- a/README.md
+++ b/README.md
@@ -36,10 +36,38 @@ Interested in learning more about the basics of virtual environments in Python
 [article](http://www.cs.allegheny.edu/sites/gkapfham/programming/research/idea/2017/07/14/Virtual-Environments/)
 to further develop your understanding of this topic.
 
+GatorGrouper requires user to use Python3. You can type `$ python --version` to check
+the current Python version on your devices. If you do not have the right Python version,
+or you can not download or upgrade Python on your devices, you can download
+[Pyenv](https://github.com/pyenv/pyenv) tool to set up the virtual environment of the
+newest Python version. 
+
+To install Pyenv, you can use [Pyenv Installer](https://github.com/pyenv/pyenv-installer).
+By typing the command in terminal:
+
+```
+$ curl https://pyenv.run | bash
+```
+
+You will successfully install Pyenv. Please make sure that you have the following lines
+in your `~/.bashrc` file.
+
+```
+export PATH="~/.pyenv/bin:$PATH"
+eval "$(pyenv init -)"
+eval "$(pyenv virtualenv-init -)"
+```
+
+Once Pyenv is installed, you can install Python 3.6.7 for your Pyenv after restart your
+terminal window by typing `pyenv install 3.6.7` and `pyenv global 3.6.7` to install.
+
+After you finish this process, you should be able to type `$ python --version` and get
+`Python 3.6.7` in your terminal.
+
 ## Initial Setup
 
 Ensure that you have installed gspread and oauth2client installed in the root
-directory of the repository.  In the terminal use the command:
+directory of the repository. In the terminal use the command:
 
 ```shell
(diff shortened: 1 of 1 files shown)


gkapfham:
Hello @Lancasterwu, thanks for creating this PR! I like what you have started and I think that you should continue to work on this. With that said, you need to receive approval from the technical leads, me, and at least one of the team leaders. Who can be the team leaders who can review these changes? We need five reviewers to merge a PR.

Also, before I start a formal review of this PR, I should make a few quick points:

* I think that we might want to emphasize 3.6.8 and 3.7.2 as these are the most recent versions of Python 3.6 and 3.6. What do others think about the appropriate version to pick?
* A person might not be using BASH, and thus we might need to stipulate that these are suggestions and that there are other options as well.
* Finally, we should also have a link to the Pyenv web site.

References:

https://www.python.org/downloads/

README.md
```diff
@@ -36,10 +36,39 @@ Interested in learning more about the basics of virtual environments in Python
 [article](http://www.cs.allegheny.edu/sites/gkapfham/programming/research/idea/2017/07/14/Virtual-Environments/)
 to further develop your understanding of this topic.
 
+GatorGrouper requires user to use Python3. You can type `python --version` to check
+the current Python version on your devices. If you do not have the right Python version,
```
Michionlion: `Python3` should be `Python 3` -- you're not referring to the executable, but the version of Python.
`devices` should not be plural.
`user` should be plural.

README.md
```diff
@@ -36,10 +36,39 @@ Interested in learning more about the basics of virtual environments in Python
...
+or you can not download or upgrade Python on your devices, you can download
+[Pyenv](https://github.com/pyenv/pyenv) tool to set up the virtual environment of
```
Michionlion: Add `the` before `Pyenv`.
Change `the` to `a` before `virtual environment`.
Change `of` to `for` before `the newest`.

README.md
```diff
@@ -36,10 +36,39 @@ Interested in learning more about the basics of virtual environments in Python
...
+the newest Python version.
+
+To install Pyenv, you can use [Pyenv Installer](https://github.com/pyenv/pyenv-installer).
```
Michionlion: This should be one sentence.

README.md
```diff
@@ -36,10 +36,39 @@ Interested in learning more about the basics of virtual environments in Python
...
+You will successfully install Pyenv. Please make sure that you have the
+following lines in your `~/.bashrc` file.
+
+```shell
+export PATH="~/.pyenv/bin:$PATH"
+eval "$(pyenv init -)"
+eval "$(pyenv virtualenv-init -)"
+```
+
+Once Pyenv is installed, you can install Python 3.6.7 for your
```
Michionlion: This should perhaps use `a version` instead of a specific version, or something similar -- the latest version of Python might change in the future!

README.md
Lancasterwu: Thanks for suggestions. Solved by https://github.com/GatorEducator/gatorgrouper/pull/123/commits/a13ebc6c3c772ec31b26f50ed6884d8bf742007c

README.md
Lancasterwu: Solved by https://github.com/GatorEducator/gatorgrouper/pull/123/commits/ba53f0a814a745115bf3c57d3ca2672817db2bd9

Michionlion: You should also reference the correct versions in the bash examples as well. Perhaps by just saying that the newest version number should be used to replace `3.6.7` or something similar.

Lancasterwu:
Hi @gkapfham , 
As @Michionlion suggested, I think it is smart to use `the latest version of Python` instead of the specific version because Python will update. I think as long as users are using the right version that they can run this program, it is fine. For differences of using different version of Python, we still need to solve issue #105 in order to know that.  
For

>     * A person might not be using BASH, and thus we might need to stipulate that these are suggestions and that there are other options as well.

I can add the instruction to install in macOS by using Homebrew and how to manually install it from Github. However, I feel the Pyenv repository has the instruction with a lot of details, and I feel it is probably too much to put in the README.md of Gatorgrouper, so I only put the automatic installer here. Since I have the link to their repository already, should I just point it out that people can go to the Pyenv repository for more solutions?
Thanks

aubreypc:
I agree that we should emphasize Python 3.7. Looking at the [change summary](https://docs.python.org/3.7/whatsnew/3.7.html), it contains several features which would make it easier to debug (e.g. breakpoints) and write maintainable code (e.g. static-ish typing features with 3.7's new typing library).

Lancasterwu:
@aubreypc Good point.  I think I'll put the newest python version after 3.7. Does that sounds right?

README.md
```diff
@@ -36,10 +36,39 @@ Interested in learning more about the basics of virtual environments in Python
 [article](http://www.cs.allegheny.edu/sites/gkapfham/programming/research/idea/2017/07/14/Virtual-Environments/)
 to further develop your understanding of this topic.
 
+GatorGrouper requires users to use Python 3. You can type `python --version` to check
+the current Python version on your device. If you do not have the right Python version,
+or you can not download or upgrade Python on your devices, you can download the
+[Pyenv](https://github.com/pyenv/pyenv) tool to set up a virtual environment for
```
gkapfham: Pyenv is not used to "set up a virtual environment".

README.md
```diff
@@ -36,10 +36,39 @@ Interested in learning more about the basics of virtual environments in Python
 [article](http://www.cs.allegheny.edu/sites/gkapfham/programming/research/idea/2017/07/14/Virtual-Environments/)
 to further develop your understanding of this topic.
 
+GatorGrouper requires users to use Python 3. You can type `python --version` to check
+the current Python version on your device. If you do not have the right Python version,
```
gkapfham: I'm not sure that "devices" is the best word. I wonder if "workstation" would be better? Or "development environment"? I think that "device" is too broad of a term.

README.md
```diff
@@ -36,10 +36,39 @@ Interested in learning more about the basics of virtual environments in Python
 [article](http://www.cs.allegheny.edu/sites/gkapfham/programming/research/idea/2017/07/14/Virtual-Environments/)
 to further develop your understanding of this topic.
 
+GatorGrouper requires users to use Python 3. You can type `python --version` to check
```
gkapfham: Where do I type this command?

README.md
```diff
@@ -36,10 +36,39 @@ Interested in learning more about the basics of virtual environments in Python
...
+the current Python version on your device. If you do not have the right Python version,
+or you can not download or upgrade Python on your devices, you can download the
```
gkapfham: See my previous comment about "device" and "devices". What is the best word @Lancasterwu and @Michionlion?

README.md
```diff
@@ -36,10 +36,39 @@ Interested in learning more about the basics of virtual environments in Python
...
+the newest Python version.
+
+To install Pyenv, you can use [Pyenv Installer](https://github.com/pyenv/pyenv-installer)
+by typing the command in terminal:
+
+```shell
+curl https://pyenv.run | bash
+```
+
+You will successfully install Pyenv. Please make sure that you have the
```
gkapfham: The "you will successfully install" does not communicate the correct point. Is it more like "This step will then install Pyenv"?

README.md
```diff
@@ -36,10 +36,39 @@ Interested in learning more about the basics of virtual environments in Python
...
+following lines in your `~/.bashrc` file.
```
gkapfham: Note that `.bashrc` is one type of file. Others may have a `.zshrc` or other type of file. You can use this as an example. But, I think that somehow, you should generalize this. Finally, while I don't think that we should really address this in the documentation, add this code slows down the startup of a shell &mdash; and so, I wrote my own code to speed up this process, which you can see in my dotfiles repository. There is a lot of content online about how to speed up this kind of code (or, similar code for Ruby or Node). Just a thought, but not something that we have to consider!

README.md
```diff
@@ -36,10 +36,39 @@ Interested in learning more about the basics of virtual environments in Python
...
+
+```shell
+export PATH="~/.pyenv/bin:$PATH"
+eval "$(pyenv init -)"
+eval "$(pyenv virtualenv-init -)"
+```
+
+Once Pyenv is installed, you can install the latest version of Python
+for your Pyenv after restart your terminal window by typing `pyenv install 3.6.7`
+and `pyenv global 3.6.7` to install.
```
gkapfham: Should the local approach also be documented?

README.md
```diff
@@ -36,10 +36,39 @@ Interested in learning more about the basics of virtual environments in Python
...
+
+After you finish this process, you should be able
+to type `python --version` and get `Python 3.6.7` in your terminal.
+
```
gkapfham: Maybe 3.6.8 and 3.7.2?

everitt-andrew: This recommendation has resulted in changes found in [this](https://github.com/Lancasterwu/gatorgrouper/pull/2) pull request.

Lancasterwu:
Hi @gkapfham @Michionlion @sutterj @toccinAC, I think my team made a lot of progress. We think it is better comparing to the previous version. Any new suggestions?

gkapfham:
Hi @Lancasterwu, can you confirm that you made all of the requested changes?

gkapfham:
Finally, @Lancasterwu, could I suggest that next time you write your commit messages so that they clearly explain the way in which, for instance, you are updating the `README.md` file? Thanks!

Lancasterwu:
Hi @gkapfham, I will check again during the weekend. Then I will make a review question to you, technical leads, and one other team leader. Also, I agree I should make my commit message to be more clear. I will do in the future.

README.md
Lancasterwu: This has been fixed

Lancasterwu:
Hi @gkapfham, I just checked all requested changes have been followed except this one:

> Should the local approach also be documented?

Can you please explain a little bit more what do you mean by `local approach`? Do you mean manually download and then set the path to it? Thanks

Michionlion:
Hi @Lancasterwu, just a note for using github: if you include the words `fixes #issuenum` in your PR description, when this PR is merged, `#issuenum` will automatically be closed and moved to the Done group in the project.

gkapfham:
Hi @Lancasterwu can you respond to the question that @Michionlion posed in an issue: does your pull request include instructions for installing Python on Windows? Is there a quick solution where we simply link to a site like this:

https://www.ics.uci.edu/~pattis/common/handouts/pythoneclipsejava/python.html

Or, alternatively, can we link to the downloads page?

https://www.python.org/downloads/windows/

I think that once we resolve this small problem then we can merge the PR. Agreed @Michionlion?

Lancasterwu:
@gkapfham Hi, we have the link of the download page in the readme already.

Lancasterwu:
@Michionlion That is an excellent suggestion, thanks!

gkapfham:
Hello @Lancasterwu, @toccinAC has been sick recently and unable to review this PR. If @barrezuetai or some other individual can review this and approve the merge, then we can merge it into the master branch right away. If that is not possible, then I suggest that we merge this without five reviews. If @barrezuetai or another team leader cannot review and approve soon, then I suggest that @aubreypc or @Michionlion go ahead and merge this PR today. Thanks!

Issue #96, opened 2019-01-29
Restructure Project so That Modules are in a Separate Directory


gkapfham:
Currently, the main modules (i.e., Python files) are in the main directory for the project. However, this makes the main directory more cluttered. This application should be significantly restructured so that all of the Python modules are moved into a new package represented by a directory with a suitable name.

aubreypc:
I have a proposal for the structure of the project:

- tests/
- src/
  - gatorgrouper.py
  - models/
    - group/
        - the existing `group_*.py` files, without the `group_`
  - views/
  - controllers/
        - the existing code for connecting to external services & API

This uses a [model - view - controller](https://en.wikipedia.org/wiki/Model%E2%80%93view%E2%80%93controller) design pattern. Roughly, models are where we define structures and functions to manage data, views are where we make different representations of the data, and controllers control the flow of data through the application and connect it to external services.

The MVC design pattern is used by many web frameworks (e.g. Django), so adopting it would make #94 more convenient if we decide to go ahead with that.

gkapfham:
One quick comment based on the suggestion by @aubreypc: to learn more about the model-view-controller approach, you should also read Chapter 32 in the _Exercises in Programming Style_ book. The title of this chapter is "Trinity" and it explains the three entities that @aubreypc mentions. If we all read this chapter, then it will help us to better discuss the trade-offs of this approach.

One last quick note: if we reorganize the system, we will need to make sure that the existing test cases can still find the modules that they are designed to test. We may also need to change the imports at the top of many of the existing files. 

Finally, do others agree that this is a priority?

aubreypc:
Here's an updated suggestion for the structure now that we are going ahead with Django -- how does this look, @quigley-c?

- tests/
- src/
  - models/
    - group/
      - group.py
      - modules for the grouping algorithms (or do these belong in controllers?)
    - user.py
    - course.py
    - assignment.py
  - views/
  - controllers/
  - static/
    - css/
    - js/
    - img/
  - templates/
- manage.py (Django's command-line interface)
- gatorgrouper.py (our command-line interface, if we choose to continue supporting this)

quigley-c:
Update on the structure after our discussion:
- tests/
- src/
    - models/
        - user.py
        - course.py
        - assignment.py
    - views/
    - controllers/
    - static/
        - css/
        - js/
        - img/
    - templates/
- manage.py
- gatorgrouper.py

barrezuetai:
[...]

PR #142, opened 2019-02-17
Redesigning with django framework

barrezuetai:
This branch took all of the existing files and moved them into the Django framework. All of the python files associated with GatorGrouper were placed in the utils file with an `__init__.py` file so that it could be called by other Django files. We'll still need to bundle of some of the programs in that directory so that it's more manageable. 

I took the test directory and put it in the GatorGrouper app directory. We'll later have to change it around so it can better work with Django's tests.py file. 

My group designed the database, but we'll still probably have to do some changes. We haven't implemented the traits table. After tomorrow's meeting with the interface team, we'll know better what route we should take when designing it. The current database model can be found in `gatorgrouper_django/gatorgrouper/models.py`.

I'm tagging @quigley-c @ilikerustoo @aubreypc @paladp @szklenskij so you guys can take a look at the redesign and offer any tips in better changing it.


gatorgrouper_django/gatorgrouper/models.py
```diff
@@ -0,0 +1,63 @@
...
+    def __str__(self):
+        return '{}'.format(self.email)
+
+
+class Semester_Class(models.Model):
+    professor_id = models.ForeignKey(
+            Professor,
+            on_delete=models.CASCADE,
+            )
+    semester = models.CharField(max_length = 50)
```
aubreypc: Do we need to check that semester strings have a consistent format before saving to the DB? Or will it be okay if some semester strings have different formats?

gatorgrouper_django/gatorgrouper/models.py
```diff
@@ -0,0 +1,63 @@
...
+    class_id = models.AutoField(primary_key=True)
+    department = models.CharField(max_length = 10)
+    class_number = models.CharField(max_length = 10)
+    class_section = models.CharField(max_length = 10)
```
aubreypc: As discussed in our meeting, I would suggest adding an attribute for a domain name so that (eventually) anyone with the link can join the class when signed in with a matching email address

gatorgrouper_django/gatorgrouper/models.py
```diff
@@ -0,0 +1,63 @@
...
+    def __str__(self):
+        return '{}: {}*{}'.format(self.department, self.class_number, self.class_section)
+
+
+class Assignments(models.Model):
+    class_id = models.ForeignKey(
+            Semester_Class,
+            on_delete=models.CASCADE,
+            )
+    assignment_id = models.CharField(max_length = 20, primary_key = True)
```
aubreypc: @chenc-allegheny's assignment creation form template contains a text input for an assignment description, so I'd suggest adding an attribute for that (probably tweet length ish? idk)

gatorgrouper_django/gatorgrouper/migrations/0001_initial.py
```diff
@@ -0,0 +1,68 @@
+# Generated by Django 2.1.7 on 2019-02-15 21:15
```
aubreypc: Since the Django script can generate this file automatically, we don't need to commit it IMO

gatorgrouper_django/gatorgrouper/models.py
barrezuetai: I was thinking of changing this and making semesters their own db. That way there's a more consistent way of saving them into the DB. Idk though

paladp:
I'm new to using Django, so I was just curious about why files like group_random, group_rrobin, group_scoring, etc., are in 'utils/'?

barrezuetai:
> I'm new to using Django, so I was just curious about why files like group_random, group_rrobin, group_scoring, etc., are in 'utils/'?

It's so that views.py can use it later when loading pages. @paladp

quigley-c:
@barrezuetai The way I understand your implementation is this way:

- root/
    - django/
        - grouper/
            - migrations/
                - django implementation with python
            - templates/
                - gatorgrouper
                    - index.html
            - tests/
            - utils/
                - gatorgrouper .py files as we know them
            - admin.py
            - apps.py
            - models.py
            - tests.py
            - urls.py
            - views.py
        - manage.py
        - db.sqlite3

Regarding #96 what is the closest way we can merge these ideas? Where do the `static/` and `controllers/` directories go? I do not think there are any major conflicts here, but I am unfamiliar with the django framework.

barrezuetai:
> @barrezuetai The way I understand your implementation is this way:
> 
>     * root/
>       
>       * django/
>         
>         * grouper/
>           
>           * migrations/
>             
>             * django implementation with python
>           * templates/
>             
>             * gatorgrouper
>               
>               * index.html
>           * tests/
>           * utils/
>             
>             * gatorgrouper .py files as we know them
>           * admin.py
>           * apps.py
>           * models.py
>           * tests.py
>           * urls.py
>           * views.py
>         * manage.py
>         * db.sqlite3
> 
> 
> Regarding #96 what is the closest way we can merge these ideas? Where do the `static/` and `controllers/` directories go? I do not think there are any major conflicts here, but I am unfamiliar with the django framework.

Static will get it's own directory in the grouper directory, and I don't think there needs to be a separate controller directory in Django, the gatorgrouper logic will be done in views.py.

aubreypc:
@quigley-c @barrezuetai: it looks like the [Django documentation describes](https://docs.djangoproject.com/en/2.1/faq/general/#django-appears-to-be-a-mvc-framework-but-you-call-the-controller-the-view-and-the-view-the-template-how-come-you-don-t-use-the-standard-names) the framework as using more of a "Model - Template - View" pattern than a "Model - View - Controller" one. In that case, Isaac may have a good point about controllers being unnecessary. If we are going to do away with controllers, I think we need to give some thought to where it makes the most sense to put our group formation code, as well as the upcoming authentication code. I suggest we do some research on MVC design patterns in general, as well as Django project structuring more specifically.

That said, aside from the decision about controllers, I would still suggest that this branch adopt the conventions discussed in #96 -- in particular, removing the `django/` nesting, and making directories for models and views rather than single files:

* tests/
* src/ (or grouper/ -- any thoughts on naming conventions?)
  * models/
    * user.py
    * course.py
    * assignment.py
  * views/
    * urls.py
  * controllers/ (?? TBD)
  * static/
  * templates/
* manage.py
* gatorgrouper.py

barrezuetai:
@aubreypc 
I'm really hesitant about changing the models.py file. We'd have to change a lot of the other files if we do that, and I'm not sure if Django would like that either. models.py came with the framework, and that's where it looks when creating the DB.

Also, from what I've read, people recommend creating a directory (can be called whatever: utils/libs/src/management), and importing the code that's found there into our views.py file. The `views.py` file is essentially where we will be loading the pages and executing all of the logic that comes with doing that i.e running the grouping program.

@sutterj What are your thoughts?

sutterj:
@aubreypc @barrezuetai 
I think that it is best for `models.py` to remain where it is and for the models (schema items for the database) themselves to remain in that file.  Django automatically configures the directory structure and most of the files when it is initialized.  It relies on these files being where it put them in order to work correctly.

gkapfham:
Hello @barrezuetai, I am wondering whether or not these changes will ultimately prevent GatorGrouper from also being used as a command-line application. My hope is that it will be possible to run it in the command-line while also interacting with it through a Django-based web interface. Is that also your plan?

gatorgrouper_django/gatorgrouper/models.py
sutterj: Django uses a single DB that contains different tables using different schema.

The format of the data that you enter can be specified in the model using, in this case, CharField and EmailField, but you can use any of the other types supported by Django and it will make sure the data is of that type before allowing you to continue.  It will not allow you to enter something that is the wrong type or format.  However, seeing as there will be users from different institutions not having a strict check on the data might be the best route.

aubreypc:
@barrezuetai @sutterj I'm a little concerned about having every model in the same file becoming unwieldy, especially if we ever need to define methods on the model classes -- but it's possible that I could be overestimating the extent to which it would be a problem. As for the question of configuration, couldn't we do:

```
grouper/
    models/
       __init__.py
       example.py
    models.py


# in models.py

# make all of example.py part of this namespace so that Django can import it from models.py like it's expecting
from grouper.models.example import *
```

gatorgrouper_django/gatorgrouper/models.py
barrezuetai: > Hello @barrezuetai, I am wondering whether or not these changes will ultimately prevent GatorGrouper from also being used as a command-line application. My hope is that it will be possible to run it in the command-line while also interacting with it through a Django-based web interface. Is that also your plan?

I didn't realize that we still wanted to keep it a command-line application along with a web app. In that case, it would change it so that it can't be run from the command-line. I'm not sure how we can make it both besides copying the code and placing one outside the Django directory and keeping another copy inside...

barrezuetai:
> Hello @barrezuetai, I am wondering whether or not these changes will ultimately prevent GatorGrouper from also being used as a command-line application. My hope is that it will be possible to run it in the command-line while also interacting with it through a Django-based web interface. Is that also your plan?

@gkapfham I didn't realize that we still wanted to keep it a command-line application along with a web app. In that case, it would change it so that it can't be run from the command-line. I'm not sure how we can make it both besides copying the code and placing one outside the Django directory and keeping another copy inside...

sutterj:
@gkapfham 
The program would still be runnable from the command line, but it would just end up a few directories deep and not on the top.  This isn’t exactly ideal, but I feel like this tool would get better use as a web app so it may make sense to go in favor of that.

sutterj:
@aubreypc 
The purpose of the models.py file is strictly to create the schema for the tables in the database.  Any methods that you would define would be located elsewhere.

gatorgrouper_django/gatorgrouper/models.py
aubreypc: @barrezuetai Using Field.choices might work? https://docs.djangoproject.com/en/2.1/ref/models/fields/#django.db.models.Field.choices

aubreypc:
@sutterj Not sure if I'm understanding your suggestion -- according to the documentation, [methods can be defined directly on models](https://docs.djangoproject.com/en/2.1/topics/db/models/#model-methods) in order to give functionality to particular rows in the table, and [methods can be defined on model managers](https://docs.djangoproject.com/en/2.1/topics/db/managers/#django.db.models.Manager) in order to provide table-wide functionality.

gkapfham:
Okay, @sutterj, I would prefer to use a command-line application over a web application. Perhaps someone could illustrate how the command-line version would change if the program becomes a web app?

sutterj:
@aubreypc 
I was referring to the state of the current file and the fact that the gatorgrouper methods are defined elsewhere.  Django can run a stand-alone python program stored in a directory.  You don’t have to load up the models.py file.

gatorgrouper_django/gatorgrouper/models.py
barrezuetai: @aubreypc I forgot about those! They'll definitely make things easier. I'll integrate them into some of the attributes tonight 💯

gatorgrouper_django/gatorgrouper/models.py
barrezuetai: Just added a description attribute with a 250 char max length. 👍

gatorgrouper_django/gatorgrouper/models.py
barrezuetai: Added a domain_name attribute! With a max length of 100 👍

gatorgrouper_django/gatorgrouper/models.py
barrezuetai: Added choices (Spring, 2019 and Fall, 2019) to the Semester_Class class to standardize responses.

gatorgrouper_django/gatorgrouper/migrations/0001_initial.py
barrezuetai: Added gatorgrouper_django/gatorgrouper/migrations/ to .gitignore

Michionlion: This file is still not removed from the actual PR, however -- you need to unstage/remove it and push.

gatorgrouper_django/gatorgrouper/templates/gatorgrouper/index.html
```diff
@@ -0,0 +1,17 @@
+<!DOCTYPE html>
+<html lang="en" dir="ltr">
+  <head>
+    <meta charset="utf-8">
+    <title>Hi</title>
+  </head>
+  <body>
+    Hi
+
+    asdfadf
```
Michionlion: This file should have some reasonable HTML stubs.

aubreypc: The add to @Michionlion's suggestion, it might be easiest to delete index.html and merge in #143 to get the design team's template drafts

aubreypc:
New schema changes look good!! 👍 
To follow up on my earlier comments: I want to get the structure of our project as finalized as we can on this first merge, since it'll be hard to adjust while lots of features are being added. Requesting @quigley-c's review on this PR since his team has been working on restructuring.

Here are my main 3 points of feedback before we merge:

- I think the Django project should be in `gatorgrouper` rather than nested in `gatorgrouper_django/`
- The `gatorgrouper/` and `gator_grouper/` should be renamed / restructured so that the difference between them is more obvious. It seems like `gator_grouper/` is for Django configuration files, maybe we could name it `config/`?
- I would still prefer to split up models.py and views.py into their own directories, but I will approve the PR even if those files remain the same.

sutterj:
@aubreypc 
You can't rename these directories without renaming the app or the project.  When you run `startproject` and `startapp`, you have to provide a name.  These are then the names that it gives to the directories.

I really would recommend against splitting up the models and views files.  In the case of this tool, I don't believe that these files are going to become so long as to end up being unmanageable (_maybe_ 200 line at the max).  Usually with Django, if there is too much going on in the models, you create another app rather than additional model files.  However, creating another app will not really work for this tool.  It will be much simpler to leave them as they are, but you can change them if you really feel that it is necessary.

sutterj:
@gkapfham 
If done correctly, it would be possible to have the same files runnable from both the web interface and the command line.  Django just provides the ability to put a different "face" onto a python program.

gatorgrouper_django/gatorgrouper/migrations/0001_initial.py
barrezuetai: Deleted File

PR #143, opened 2019-02-17
I-135-bootstrap - interface

ElisiaW:


gkapfham:
I agree with @aubreypc that there is no need to include the README file for Bootstrap in the repository for GatorGrouper.

aubreypc:
@Michionlion Good point about committing the minified files!

One justification for self-hosting our CSS rather than using a CDN would be that, as described in #144, we will be able to customize the vanilla Bootstrap styles if need be in order to fit our changing design goals over the next few weeks. That said, if others don't see this as being a priority, then we can just get our Bootstrap CSS and JS from a CDN as you suggest. @GatorEducator/ui-ux-web-design-team: any thoughts?

Michionlion:
> @Michionlion Good point about committing the minified files!
> 
> One justification for self-hosting our CSS rather than using a CDN would be that, as described in #144, we will be able to customize the vanilla Bootstrap styles if need be in order to fit our changing design goals over the next few weeks. That said, if others don't see this as being a priority, then we can just get our Bootstrap CSS and JS from a CDN as you suggest. @GatorEducator/ui-ux-web-design-team: any thoughts?

Just a comment on that: you can still modify the Bootstrap CSS by overriding what you want to change, there is still no need to keep it locally.

Michionlion:
Hi @ElisiaW, this PR is attempting to merge into the old Django branch, which no longer exists and has had major changes. You need to rebase your branch off of either the `develop` branch or the `master` branch after `develop` has been merged. Then, you can merge your branch to master. Currently this PR is very large, and will likely need to be split up before it will be approved.

I'm inclined to close this PR; discussion about design should be on issues rather than a PR.

ElisiaW:
Thanks for the feedback @Michionlion. I just rebased the branch to `master`, as it would give a fatal error when trying to rebase it to `develop`.

Michionlion:
> Thanks for the feedback @Michionlion. I just rebased the branch to `master`, as it would give a fatal error when trying to rebase it to `develop`.

You definitely need to incorporate changes from `develop`, however, since your branch requires Django, and `master` does not have the current implementation or restructure. This could happen later on, however. A new PR will also need to be created when you are ready to merge.

chenc-allegheny:
I'm closing this because @Michionlion and I moved the branches from @ElisiaW's fork to this repository, and it needs to be recreated.

PR #150, opened 2019-02-19
Fix some details in README file and add pipenv, pylint and flake8 integration

### LINKED ISSUE #95: Use Pipenv for Dependency Management
This project currently uses Pip for dependency management. However, this leads to the development and application dependencies currently being mixed in the same `requirements.txt` file. The project should be overhauled so that it now uses Pipenv instead and then the documentation should have new instructions for installing the dependencies and running the application and its test suite.

### LINKED ISSUE #99: Report Coverage Through PyTest and to CodeCov.io Instead
The README contains instructions for running code coverage analysis that may not be standard. First, all of the dependencies should be installed with Pipenv. Second, coverage should be collected exclusively through a run of Pytest. Finally, details about reporting coverage to Coveralls do not need to be in the README file as they are instructions exclusively for use in Travis CI. I also suggest that we adopt CodeCov.io instead of using Coveralls.

### LINKED ISSUE #100: Delete Travis CI Integration Documentation
The README file currently explains how to enable Travis CI. However, this is not needed in the README and thus I suggest that it be removed entirely. You should also ensure that the Travis CI integration specified in the `.travis.yml` file is up-to-date and working in accordance with the newest version of travis-ci.com.

### LINKED ISSUE #101: Add Badges to the Top of the README
The project currently has a badge at the very bottom of the README. This should be moved to the top. Also, the integration should be with CodeCov.io instead of Coveralls. Finally, you should add extra badges that report, for instance, the status of the build.

### LINKED ISSUE #103: Reconfigure Travis CI so That is Uses Pipenv
The Travis CI configuration given in the README file is out of date in regards to the current standards for using Travis CI. This content should also be thoroughly revised so that it uses Pipenv for dependency management.

### LINKED ISSUE #104: Adopt Better Standards for Project Code Linting

### LINKED ISSUE #105: Test with Multiple Versions of Python Using Pyenv and Tox

### LINKED ISSUE #124: Add Flake8 and Pylint and Other Linting Tools Directly to Travis CI
There are 2 lines of `tests/test_flake8.py` was not covered. Also, Gatorgrouper is using `coveralls.io` to show the code coverage, should we change it to `codecov.io` so we can easily know which line wasn't covered?

huangs1:
This PR:
- fixes #95 
- fixes #99
- fixes #100 
- fixes #101 
- fixes #103 
- fixes #104 
- fixes #105 
- fixes #124 

@huangs1 
-
#100 Delete Travis CI Integration in README file

@yeej2
-
#101 Move the original badge (which was also changed from coveralls to codecov) with some additional new badges (build status and language) to the top of the README file and edit details of `Activation Coveralls` in README file

@ilikerustoo 
-
#95 and #103 Add `Pipenv` integration, delete `requirements.txt`, add some dependencies in `Pipfile`
#99 Add pytest codecov integration, add some `dotfiles` to aid code coverage.
#105 Add multiple versions of python in Travis file and tox was deemed unnecessary for now.
#124 and #104 Add `pylint` and `flake8` linting tools to `.travis.yml` 
However, running `black` caused `bad-continuation` `pylint` issues in `gatorgrouper.py` and `parse_arguments.py`. So, @ilikerustoo suppressed/disabled issues in the aforementioned python modules. Also, he added a check in `Travis` to see if `black` was running on files and modules.


.travis.yml
```diff
@@ -1,31 +1,55 @@
...
+# --> run the test suite with pytest
+# --> lint the functions with flake8
+# --> lint the test suites with flake8
+# --> lint the functions with pylint
+# --> lint the test suites with pylint
+# --> lint the README documentation with mdl
 script:
-  - pytest tests
+  - pipenv run pytest tests --cov-config pytest.cov --cov
+  - pipenv run black *.py --check
```
Michionlion: This probably doesn't explore recursively down directories, so when the final refactored versions are added, this will also need to be changed. Could there be a better way?

README.md
```diff
@@ -282,43 +286,50 @@ coverage report
 ```yml
```
Michionlion: This section does not seem relevant -- you should follow the advice given in #100 and fully delete this information.

README.md
```diff
@@ -282,43 +286,50 @@ coverage report
...
+  - pipenv run flake8 *.py
+  - pipenv run pylint tests
+  - pipenv run pylint *.py
   - mdl README.md
-  - coverage run --source tests -m py.test
-  - coverage report
 ```
 
-## Activating Coveralls
+## Activating CodeCov
```
Michionlion: Again, this is not relevant in the README. Perhaps some additional documentation elsewhere on how to use codecov or travis would be useful, but it does not belong in the readme, since everyone who uses this tool does not need to do this.

rebalance.py
```diff
@@ -7,22 +7,24 @@ def rebalance(student_groups, scores, ave):
     """rebalances the groups of students"""
 
     # average score of the groups
-    ave = int(ave / len(scores))
+    ave = int(ave // len(scores))
```
Michionlion: This change could make a difference in actual use (using int division instead of floating point). Additionally, if int division is used, casting to int is redundant. Is this well tested?

.travis.yml
ilikerustoo: Yes I'm waiting for the restructured to be finalized to make the appropriate changes.

rebalance.py
ilikerustoo: I could not test this since `Gatorgrouper` is not working correctly but I am going to work with @Lancasterwu to resovle this. An alternative to is to revert the changes and import  `__future__statement`

README.md
huangs1: I have deleted this section from the README.

README.md
huangs1: I have deleted this section from the README.

rebalance.py
ilikerustoo: Reverted the changes from // to / and disabled the old-division error locally as I found this was not working when included in the `.pylintrc`. Our team will be working with @Lancasterwu to make test cases and code coverage

Michionlion:
Closing this PR as #154 merges these changes.

Issue #146, opened 2019-02-18
Store secret keys securely


aubreypc:
Before eventually deploying our web application, we will want to find a way to keep any secret keys out of the repository -- perhaps with environment variables? -- so that nothing sensitive becomes publicly visible on GitHub.

ilikerustoo:
I believe I figured out the way to do this. Please see my branch [hide-secret-key](https://github.com/GatorEducator/gatorgrouper/tree/bugfix/hide-secret-key)

szklenskij:
@ilikerustoo fixed this issued in pull request `Bugfix/fixing bugs in develop` #159 which I approved for merge

PR #140, opened 2019-02-15
Removes GoogleSheet Integration and Updates README to Provide Tradeoffs

paladp:

PR #154, opened 2019-02-20
Merging Develop Branch To Master

### LINKED ISSUE #94: Creating Frontend Web Application to Increase Usability
I believe many faculty at Allegheny would find this software useful but may find the process of pulling the repo, downloading the proper dependencies challenging thus dissuading them from using it. I propose building a simple web application that uses gator grouper on a python server in background. This implementation could be done in a micro framework like flask or pyramid. The front-end design could be done in a Vue, or just plain old Vanilla JS, HTML, CSS. The hosting of this app could be on a free heroku app or on Allegheny's Server for internal use.

### LINKED ISSUE #97: Delete Google Forms Interaction and Use a Local CSV File
The system currently provides a feature that supports interaction with Google Forms. While there is merit in this feature, my concern is that it adds complexity and it hard to test with an automated test suite. As such, I suggest that we remove the Google Forms integration and instead ask the user to specify a CSV file that contains the results from a Google Form. While this is extra work for the user, I judge that it would make it easier for us to deliver a production quality tool.

barrezuetai:
This branch incorporates changes made by multiple teams that fix these issues:
- issue #95 
- issue #99
- issue #100 
- issue #101 
- issue #103 
- issue #104 
- issue #105 
- issue #124 

# Implementing Django (#142)
A Django app has been added to the root of the project. It's been redesigned a bit to contain modified test files, source files, and `models.py` file. Along with the Django implementation comes its other files like `views.py` and `urls.py` and all of the configuration files. Because implementing Django changed the organization of the repository, changes to the `.travis` file had to be made as well. In order to simplify the use of this repository, we've installed a `Pipfile` so that developers could install all of the dependencies with ease. This fixes #94.

## Test Files
Because this branch restructures where the tests lie, we had to make changes to the import statements of these files so that pytest can run them without error. They now refer to code in the utils directory instead of the previous directory name.

## Source Files
The source files have been grouped together in a directory called utils, that will be used by models.py, the test cases, and views.py. Because the name of their path was changed, the import statements in these files also had to be changed for them to run.

## Models.py
This branch holds the current schema for the database in models.py. Django uses this file to create the database, and as such will be the file that needs to be modified if changes to the DB need to be made. 

## views.py and urls.py
These files will be  further used and modified when we import template files that will be displayed by the browser.

## Travis
We've reconfigured Travis so that it points to the correct location of the tests files. 

# Pipfile and more (#150)
Because we switched from having a file with dependency names to using pipenv, we've added a pipfile so that developers can use `pipenv install --dev` to create a virtual environment with all of the project dependencies.

## @huangs1 

#100 Delete Travis CI Integration in README file

## @yeej2

#101 Move the original badge (which was also changed from coveralls to codecov) with some additional new badges (build status and language) to the top of the README file and edit details of `Activation Coveralls` in README file

## @ilikerustoo 

- #95 and #103 Add `Pipenv` integration, delete `requirements.txt`, add some dependencies in `Pipfile`
- #99 Add pytest codecov integration, add some `dotfiles` to aid code coverage.
- #105 Add multiple versions of python in Travis file and tox was deemed unnecessary for now.
- #124 and #104 Add `pylint` and `flake8` linting tools to `.travis.yml` 
- #146 Hide key in PR #159 
-  In PR #156  Fix path issues, `pylint`, `flake8` errors. Disable suppression messaged in  `.pylintrc`. Remove `.coveragerc` and extra lint command for linting.
- Move `django` project one level up, removing `gatorgrouper_django` and extra `tests` folder during `django` project restructure.

# Removes Google Spreadsheet Integration (#140)
`spreadsheet.py` was removed, along with relevant implementation, to simplify command-line application and web design. Fixes #97.


.coveragerc
```diff
@@ -0,0 +1,3 @@
+[run]
+omit=tests/*, .venv/*
```
Michionlion: Should this specify `gatorgrouper/tests/*`? Also, is this for codecov or the old coverage? It may need to be deleted.

.travis.yml
```diff
@@ -1,31 +1,55 @@
...
+# --> lint the functions with flake8
+# --> lint the test suites with flake8
+# --> lint the functions with pylint
+# --> lint the test suites with pylint
+# --> lint the README documentation with mdl
 script:
-  - pytest tests
+  - pipenv run pytest gatorgrouper/tests --cov-config pytest.cov --cov
+  - pipenv run black */**.py --check
+  - pipenv run black tests --check
```
Michionlion: There are duplicate `black --check` runs here -- the first will check all the files that the second checks anyways.

Pipfile
```diff
@@ -0,0 +1,30 @@
...
+url = "https://pypi.org/simple"
+verify_ssl = true
+
+[dev-packages]
+pylint = "*"
+black = "*"
+pytest = "*"
+six = "*"
+codecov = "*"
+"flake8" = "*"
```
Michionlion: Why is flake8 in `"` here? Does it matter?

Pipfile
```diff
@@ -0,0 +1,30 @@
...
+gspread =  "*"
+oauth2client = "*"
+
+[packages]
+pathlib = "*"
+py = "*"
+django = "*"
+
+[requires]
+python_version = "3.6"
```
Michionlion: Should python 3.6 be explicitly required?

barrezuetai: @ilikerustoo

.coveragerc
ilikerustoo: I believe it is for `codecov` as omitting `.venv/*` worked for me previously. However I will look into its need and correct path if necessary.

Pipfile
ilikerustoo: I just copied over what already existed in `GatorGrader`. That is how it was but I will check if it needs to be in quotes.

Pipfile
ilikerustoo: We deleted this but it might have accidentally come back in the merging process. Also for a sure check for errors, we should be looking at `Travis` instead of individual machines.

quigley-c:
@aubreypc @sutterj addressed this concern in the other PR #142 

> @aubreypc
> You can't rename these directories without renaming the app or the project. When you run `startproject` and `startapp`, you have to provide a name. These are then the names that it gives to the directories.
> 
> ...

It sounds like it won't be possible to rename these directories. Is this incorrect?

barrezuetai:
> Thanks for the PR! A couple points of feedback:
> 
>     * The name of the `gator_grouper` package should be changed, as it is confusing to have both `gator_grouper/` and `gatorgrouper/`. Especially since there are files with the same name in both dirs, like `urls.py` -- the distinction should be obvious just from looking at the structure of the project.
> 
>     * Does the database file need to be committed to the repo, or will Django create one automatically on startup if it doesn't exist?

@aubreypc The database does need to be there cause it holds superuser info and other information

sutterj:
@aubreypc @quigley-c 
It is correct that you cannot rename these.  The outer directory is the name of the project and the inner directory is the name of the app.  You could start over in order to give them new names and then copy in the files that you changed (making sure that you change the app name where appropriate), but I don't really think that it is necessary.  The distinction between these directories and the structure of it all is an inherent part of Django.

aubreypc:
@quigley-c It is possible to change the initial `startproject` and `startapp` names, at least according to [this blog post](https://www.techinfected.net/2016/08/how-to-change-your-django-project-name.html). If that doesn't work, we could run `startproject` again and then copy in any files which were modified from that base template.

barrezuetai:
@aubreypc What would you recommend the name changes be?

sutterj:
@aubreypc @barrezuetai 
I would seriously recommend against changing them in the existing files.  There are likely to be a slew of issues so if you are going to change them, run the `startproject` and `startapp` commands again.  There are a lot of files that are created in the background that are going to have the old names on them that have the potential not to get changed.  Django is pretty finicky/particular in this area.

aubreypc:
@barrezuetai Personally I would suggest renaming `gator_grouper` to `config` to distinguish the Django project configuration from the gatorgrouper app -- does anyone else have suggestions?

sutterj:
@aubreypc 
I personally think that the name of the app and the project should somehow name the project.  I would call them something like `gatorgrouper_project` and `gatorgrouper_app`.  I would recommend not using `config` as Django uses `.config` and `AppConfig`.  You could end up with something like `config.config`.

Michionlion:
@aubreypc @sutterj @barrezuetai I would suggest looking into the idea of Django apps -- @ilikerustoo and I looked into that but it turned out to be too complicated for this restructure. This could enable renaming `gator_grouper` to `gatorgrouper` and `gatorgrouper` to `applications` or something similar.

gator_grouper/settings.py
```diff
@@ -0,0 +1,116 @@
...
+
+# Build paths inside the project like this: os.path.join(BASE_DIR, ...)
+BASE_DIR = os.path.dirname(os.path.dirname(os.path.abspath(__file__)))
+
+
+# Quick-start development settings - unsuitable for production
+# See https://docs.djangoproject.com/en/2.1/howto/deployment/checklist/
+
+# SECURITY WARNING: keep the secret key used in production secret!
+SECRET_KEY = "jc*l8zi5+h1_^f6i)x@0v%l=rw=219!i(dyf@2%_ez1e6mta$x"
```
Michionlion: Very bad thing here!!!!

This needs to be taken care of and possibly changed since we absolutely *cannot* publish a secret key.

If this can be regenerated (@sutterj?), then we should merge this and then create a new PR to handle this issue. If not, other methods should be looked into.

barrezuetai: http://fearofcode.github.io/blog/2013/01/15/how-to-scrub-sensitive-information-from-django-settings-dot-py-files/

I can follow this and change it right now

Michionlion: Note that we will still need to regenerate the key, and that this must be done after this merge is completed, according to the [git rewriting instructions](https://help.github.com/en/articles/removing-sensitive-data-from-a-repository).

sutterj: If you are planning on rerunning startapp and startproject in order to rename the directories (without the process of using a Django app) then you will end up getting a new key and could delete the old one.

ilikerustoo:
> @aubreypc @sutterj @barrezuetai I would suggest looking into the idea of Django apps -- @ilikerustoo and I looked into that but it turned out to be too complicated for this restructure. This could enable renaming `gator_grouper` to `gatorgrouper` and `gatorgrouper` to `applications` or something similar.
> 
> There's no reason not to be able to rename the folders, as they are all just specified in the generated python files.

@Michionlion I will look into it. Although I am not familiar with `django`, it should not be too much of a problem to rename some things. We need to make this project easy to understand so that the people who work on `Gatorgrouper` after us do not run into too many problems. @aubreypc I agree with the name change and discussed it with @barrezuetai and @Michionlion a few days ago. This was part of my reasoning for getting rid of the `gatorgrouper_django` root folder.

gator_grouper/settings.py
ilikerustoo: @Michionlion agreed. @barrezuetai  I can help with this.

.travis.yml
```diff
@@ -1,31 +1,54 @@
+dist: xenial
 language: python
 python:
+  - "3.7"
+  - "3.6"
   - "3.5"
```
ilikerustoo: @aubreypc `python type hinting` is only supported in versions `3.6` and above. Would you like me to get rid of versions `3.6` and below?

gatorgrouper/apps.py
```diff
@@ -0,0 +1,8 @@
+""" This is undocumented """
+from django.apps import AppConfig
+
+
+class GatorgrouperConfig(AppConfig):
+    """ This is undocumented """
+
+    name = "gatorgrouper"
```
ilikerustoo: @barrezuetai  @sutterj Will changing `name=gatorgrouper` here allow us to rename the app folder to something else?

gatorgrouper/models.py
```diff
@@ -0,0 +1,71 @@
+""" This is undocumented """
```
ilikerustoo: @aubreypc @barrezuetai  There are several `""" This is undocumented """` that I added to pass travis and `pylint`. Whoever has more understanding of these `django` files should change these into something more appropriate. This is not the only file with these `docstrings`.

gatorgrouper/utils/display.py
```diff
@@ -8,15 +8,15 @@ def display_student_groups(student_groups):
...
         for student in student_group:
             place_counter += 1
-            #print(student[0])
-            print('\033[0m' + student[0])   # converts the students back to original font
+            # print(student[0])
+            print("\033[0m" + student[0])  # converts the students back to original font
         print()
         group_counter = group_counter + 1
-    logging.info("Found " + str(place_counter) + " students")
+    logging.info("Found %d students", place_counter)
```
ilikerustoo: @Michionlion @aubreypc . I have changed logging to pass `pylint` errors but I would like someone to look over if I have correctly used the correct data type e.g `%d`. Additionally as I discussed with @gkapfham, logging may be unnecessary if we have good test cases and may be deleted. However, this topic does not take priority and can be done later if need be.

Lancasterwu:
There is an empty `test` folder in the root of the repository which does not make any sense. I will go ahead and delete it.

Lancasterwu:
The `students.csv` file seems necessary for now. The testing team (@GatorEducator/team-lancaster and @GatorEducator/team-spencer ) would not be able to run the restructured program without an input file. We should put it back, and remove it after the `json` feature is successfully implemented.

ilikerustoo:
@lancaster pleas refer to my PR #156 before making more commits. The changes you mentioned have already been fixed.

Lancasterwu:
Thanks @ilikerustoo, just want to make sure that is `bugfix/hide-secret-key` the most recent branch with everything updated? Or the correct and working branch can only be the `develop` branch after both `bugfix/hide-secret-key` and `bugfix/fixing-bugs-in-develop` merged in?

Issue #184, opened 2019-02-27
Debugging and Refactoring the Current Python Code


Lancasterwu:
With PR #180, we can finally run the restructured `gatorgrouper`.  

However, there are still some bugs and bad designed python code.

The program was only refactored in this structure, some specific functions still looks terrible.

I want to assign @GatorEducator/team-lancaster and @GatorEducator/team-spencer (except @huangs1 and @ilikerustoo) to go over all code in details to improve them. 

"The refactored code should be clean, efficient, and maintainable code that is well documented"

Please create a branch following the style `enhance/debug-improve-YOURFILENAME`
Note that you should not put `.py` after `YOURFILENAME`.

To be organized, please comment under this Issue to claim the file you are going to work.

You should be responsible for having a good understanding of the file that you are assigned and be able to explain to other people.

😄 😆 😊 😃 😏 😄 😆 😊 😃 😏 😄 😆 😊 😃 😏 😄 😆 😊 😃 😏 😄 😆 😊 😃 😏 

-------------------------------------------------------------------------------------------------------
We now have:
* @thomad74 for `read_student_file.py`
* @ALEXANDERB82  for `group_scoring.py`
* @wattob for `group_random.py`
* @everitt-andrew  for `wgroup_rrobin.py`
* @reibelj for `parse_arguments.py`
* @Lancasterwu for `display.py`
* @yeej2 for `remove_absent_students.py` and `rebalance.py`
* @baldeosinghm   for `gatorgrouper.py`

There are only two configuration files left
* `defaults.py` - short
* `constants.py` - short

-------------------------------------------------------------------------------------------------------
😄 😆 😊 😃 😏 😄 😆 😊 😃 😏 😄 😆 😊 😃 😏 😄 😆 😊 😃 😏 😄 😆 😊 😃 😏 

When ever you find a bug or bad designed code or some thing that does not make sense to you, please make a note under this issue and raise a new issue to clearly explain it.

For example: 

Comment under this issue: 

```
reibelj commented a day ago

I will look into `parse_arguments.py`

Issue #194 to fix a bug related to invalid group number
```

Sample Issue that you should raise: https://github.com/GatorEducator/gatorgrouper/issues/194

ilikerustoo:
Could to be more specific in examples of what would need improving? This is probably low priority for now since we have the application finally up and running when PR #180 merges.

baldeosinghm:
[...]

PR #185, opened 2019-02-27
Increase Code Coverage for All Functions

### LINKED ISSUE #137: Low code coverage with Codecov.io - tests needed
Right now using `CodeCov` and the following command `pipenv run pytest -x -s --cov-config pytest.cov --cov-report term-missing --cov` we are getting a total coverage of 58%.  Previously we had low coverage due the incorrect configuration of `coverall` that checked coverage on the tests themselves this providing a falsely high code coverage.

We need more test cases that cover the rest of the code and this can be achieved by simply making tests for the missing statements/line below:

````
----------- coverage: platform linux, python 3.6.7-final-0 -----------
Name                        Stmts   Miss  Cover   Missing
---------------------------------------------------------
defaults.py                     5      0   100%
display.py                     22      4    82%   24-27
gatorgrouper.py                43     26    40%   25-82
group_random.py                50     20    60%   46-69
group_rrobin.py                55     24    56%   54-87
group_scoring.py               15      0   100%
parse_arguments.py             53     12    77%   94, 103, 116-128
read_student_file.py           18     13    28%   15-29
remove_absent_students.py      10      0   100%
spreadsheet.py                 35     28    20%   14-54
---------------------------------------------------------
TOTAL                         306    127    58%

````

We must aim for at least 80% coverage.

### PR DESCRIPTION
**Pull request by Lancasterwu**
# Description of Pull Request

This PR is provided by @GatorEducator/team-lancaster and @GatorEducator/team-spencer to **increase the code coverage from 64.15% to 99.17%** with the following features:

* **Increasing code coverage for `display.py` from 82% to 100%.**
    * Provide `test_display_welcome_message` and restructure `display_welcome_message` to make it can be easier tested.  @Lancasterwu 
* **Increasing code coverage for `grouprandom.py` from 60% to 100%.**
    * Provide test functions to test `group_random_num_group`. @wattob
* **Increasing code coverage for `grouprrobin.py` from 56% to 100%.**
    * Provide test functions to test ` group_rrobin_num_group` @ALEXANDERB82 & @everitt-andrew 
* **increasing code coverage for `readstudentfile.py` from 28% to 100%.**
    * Provide `test_read_file_populates_csvdata_0` to test `read_student_file` and add `students.csv` as an input to test. @thomad74 
* **Increasing code coverage for `parseargument.py` from 77% to 98%.**
    * Provide a new file `test_num_group.py` with test functions to test `check_valid_num_group`. @reibelj 
* **Increasing code coverage for `gatorgrouper.py` from 40% to 100%.**
    * Provide `pragma` to not report test coverage for `if __name__ == "__main__"` @baldeosinghm & @yeej2

**NOTE:** 
* There are two `quit()` functions was not able to be tested in `parseargument.py`, which caused that 2% of coverage missing.
* Random test input generator `Hypothesis` still need to be implemented. @mmarconi and @libailong are currently working on that.
* This PR should only be merged in after PR https://github.com/GatorEducator/gatorgrouper/pull/180 is merged. In order to run the correct linting tool with the right path, some changes are same to that PR.

This PR Fixes #137 

## Type of Change

- [ ] Bug fix
- [ ] Breaking change
- [ ] New feature
- [ ] Documentation update
- [X] Improving Test Coverage

## Tags

@Lancasterwu @huangs1 @gkapfham @aubreypc @Michionlion

### CODE REVIEW

File: gatorgrouper/tests/test_num_group.py [hunk 1]
```diff
@@ -0,0 +1,53 @@
+"""Testing number of groups"""
```
Michionlion: This file needs another test, where you actually check a real group number. Additionally, `test_check_number_groups_size_zero` likely should get a false result from `check_valid_num_group`, since zero groups is not possible to handle. These tests should be reviewed.

gatorgrouper/tests/test_read_student_file.py
```diff
@@ -0,0 +1,7 @@
+"""Testing number of groups"""
+from utils import read_student_file
+
+
+def test_read_file_populates_csvdata_0():
```
Michionlion: Why does this method name end with `_0`?

Additionally, the `!= 0` does not actually check anything -- you should check that it contains something, perhaps by checking if it is truthy, or better yet, by constructing a `csv` file in code and then checking it reads the correct data.

This is a really bad test.

gatorgrouper/utils/display.py
```diff
@@ -21,10 +21,11 @@ def display_student_groups(student_groups):
...
 def display_welcome_message():
     """ Display a welcome message """
-    print()
-    print("GatorGrouper: Automatically Assign Students to Groups")
-    print("https://github.com/GatorGrouper/gatorgrouper")
-    print()
+    firstline = ""
+    secondline = "GatorGrouper: Automatically Assign Students to Groups"
+    thirdline = "https://github.com/GatorGrouper/gatorgrouper"
+    forthline = ""
```
Michionlion: This is ***not*** the same code as before -- was this actually tested by running it at all?

.gitignore
```diff
@@ -1,7 +1,6 @@
 __pycache__
 .cache
 *.pyc
-*.csv
 gatorgrouper_django/gatorgrouper/migrations/
```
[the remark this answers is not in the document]
Lancasterwu: Hi @Michionlion, this is due to that we do need `students.csv` for `test_read_student_file.py` as an sample file to test `read_student_file.py`. However, as you mentioned in other change requests. We probably need to write a better test function for this without use `students.csv`.

gatorgrouper/tests/students.csv
```diff
@@ -0,0 +1,28 @@
+"delgrecoj","True","True","False","True"
```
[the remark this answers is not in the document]
Lancasterwu: This is the same reason that i mentioned for the previous requested change, we will rewrite the testing function in `test_read_student_file.py`.

gatorgrouper/tests/test_display.py
```diff
@@ -12,3 +12,14 @@ def test_display_student_groups(capsys):
...
     assert out.startswith("\033[0;32m" + "\033[1m" + "\033[4m" + "Group 1")
+
+
+def test_display_welcome_message():
+    """ Test if the welcome message is correctly displayed"""
+    firstline = ""
+    secondline = "GatorGrouper: Automatically Assign Students to Groups"
+    thirdline = "https://github.com/GatorGrouper/gatorgrouper"
+    forthline = ""
+    expected_output = firstline + secondline + thirdline + forthline
```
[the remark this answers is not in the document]
Lancasterwu: Yes this test passed `pipenv run pytest -x -s --cov-config pytest.cov --cov-report term-missing --cov`.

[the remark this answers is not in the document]
Lancasterwu: @Michionlion Actually Spencer and I just tried to actually run it, and there is no output. However, we are able to fix this by adding `print()` right after `return`. 
The main reason that I need to change the structure of this is due to the old function returns `None` variable which is hard to be tested.

gatorgrouper/tests/test_group_method.py
```diff
@@ -106,6 +122,31 @@ def test_round_robin():
...
+        ["Jon", False, False, False],
+        ["Angie", False, False, False],
+        ["Izaak", False, False, False],
+        ["Jacob", False, False, False],
+    ]
+    numgrps = 4
+    response_output = group_rrobin.group_rrobin_num_group(lst, numgrps)
+    assert len(response_output[0]) == 3
+    assert len(response_output) == numgrps
+    assert (["Dan", True, True, True] in response_output[0]) is True
```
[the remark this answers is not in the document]
Lancasterwu: @everitt-andrew @ALEXANDERB82 should be responsible for these changes. I will also work on this with Spencer.

gatorgrouper/tests/test_read_student_file.py
Lancasterwu: We are going to rewrite this test function.

gatorgrouper/utils/display.py
Lancasterwu: This function can be tested by running `pytest`. However, the function itself does not have any output so we will fix that.

huangs1: I have changed it so that the function is printing out the expected output.

Lancasterwu: @Michionlion I reformatted the code in a wired way which I don't know if the program is actually being tested.  It does not make sense to me but it has the correct output and passed `pytest`.

gatorgrouper/tests/test_group_method.py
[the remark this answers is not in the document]
Lancasterwu: Hi @Michionlion, we have tried to run this test case several times and it does not create a tuple. By using the debugging print line 
```
print (["Austin", True, True, True] in response_output[1])
```
we got the output:
```
gatorgrouper/tests/test_group_method.py::test_round_robin ✓      66% ██████▋   
True
```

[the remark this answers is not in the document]
Michionlion: That does indeed seem true, but if possible it may be best to still remove the parenthesis.

gatorgrouper/utils/display.py
```diff
@@ -21,10 +21,9 @@ def display_student_groups(student_groups):
 
 def display_welcome_message():
     """ Display a welcome message """
-    print()
-    print("GatorGrouper: Automatically Assign Students to Groups")
-    print("https://github.com/GatorGrouper/gatorgrouper")
-    print()
+    firstline = "\nGatorGrouper: Automatically Assign Students to Groups"
+    secondline = "\nhttps://github.com/GatorGrouper/gatorgrouper\n"
+    return print(firstline + secondline)
```
Michionlion: `print` does not return anything -- in order to test this method, you need to capture `sys.out`, which is covered on this [page](https://docs.pytest.org/en/latest/capture.html#accessing-captured-output-from-a-test-function) of the `pytest` documentation.

Is there any reason why the code here has to change? There is no need to create string variables or anything like that, multiple print statements are perfectly fine. I would highly suggest using the original code and simply using the `capsys` fixture as described in the link above to capture the printed output.

gatorgrouper/utils/display.py
Michionlion: By tested I meant the program run at the command line -- what you should have observed is that everything was printed on one line, if at all.

gatorgrouper/utils/display.py
Lancasterwu: Thanks, we will work on this soon.

gatorgrouper/tests/test_group_method.py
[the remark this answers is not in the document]
Lancasterwu: I just changed the assertion in this function and some other old test functions that were using parenthesis.

gatorgrouper/tests/test_num_group.py
Lancasterwu: Hi @Michionlion, `test_check_number_groups_size_zero` has the correct assertion due to that `--num-group 0` is able to be handled, which it should not. I will let @reibelj to take care of this in another PR since this is the part of Issue https://github.com/GatorEducator/gatorgrouper/issues/184 and he is assigned to `parse_arguments`.
Additionally, what do you mean by "check a real group number"?

Michionlion: You should check a number like `2`, since all of your tests test exceptional conditions in some form. You should also ensure a normal condition works, and a group size of `2` or `3` would make sense for that.

Lancasterwu: Thanks, will do.

gatorgrouper/utils/display.py
Lancasterwu: This has been fixed by https://github.com/GatorEducator/gatorgrouper/pull/185/commits/eecc5d46a1fe67615f931db711e76a745cc2fc59

gatorgrouper/tests/test_num_group.py
Lancasterwu: Solved by https://github.com/GatorEducator/gatorgrouper/pull/185/commits/72ea4950e5dcae701459351ccfcdf1947beb2453

gatorgrouper/tests/test_read_student_file.py
```diff
@@ -1,7 +1,22 @@
...
+import csv
+#from coreapi import Client
 from utils import read_student_file
 
 
-def test_read_file_populates_csvdata_0():
-    """Checks that the read student file populates data from a CSV."""
-    assert read_student_file.read_student_file("gatorgrouper/tests/students.csv") != 0
+@pytest.fixture(scope='session')
+def file_name(tmpdir_factory):
```
Michionlion: This is good but the fixture should have a more specific name -- perhaps `csv_file`? Otherwise, if there are other tests that want to use file fixtures, there might be conflicts in the future.

gatorgrouper/utils/read_student_file.py
```diff
@@ -17,7 +17,7 @@ def read_student_file(filepath):
 
     # transform into desired output
     responses = list()
-    for record in csvdata:
+    for record in csvdata[1:]:
```
Michionlion: Can you explain what this change means? Why does it need to be different? That information could be added to the PR description.

gatorgrouper/tests/test_num_group.py
```diff
@@ -0,0 +1,75 @@
...
+        "michael",
+        "jacob",
+        "kapfhammer",
+    ]
+    student_groups = parse_arguments.check_valid_num_group(numgrp, student_identifiers)
+    assert student_groups
+
+
+def test_check_number_groups_size_muti():
+    """Checking 1 group"""
```
Michionlion: There's a misspelling of `multi` here. :smiley: 

Additionally, the docstring for this method is incorrect.

Lastly, it is probably not good practice to test two different group sizes in one test. This is an ***excellent*** chance to use a single parameterized test case!!

gatorgrouper/tests/test_read_student_file.py
Lancasterwu: I have changed to `generate_csv`, would that work? or `csv_name` makes more sense.

gatorgrouper/utils/read_student_file.py
Lancasterwu: This is due to the bad design of `read_student_file.py` and the original sample csv. If the input csv file has the header, that function will make the header name as the first list and there is no exception handling for it. I used `[1:]` to avoid collecting output of the first list which can be the header. However, this is not the solution for this bad design. @thomad74 is responsible for this file based on Issue https://github.com/GatorEducator/gatorgrouper/issues/184. There will be another PR to fix this bad design.

gatorgrouper/tests/test_num_group.py
Lancasterwu: The test cases are not the finalized version. @libailong and @mmarconi are responsible for implementing `Hypothesis` which can be used to write parameterized test case with randomized testing input. I think it is probably fine to leave it like that for now since it will be changed later anyway.

.gitignore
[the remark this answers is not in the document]
Lancasterwu: This has been added back

gatorgrouper/tests/students.csv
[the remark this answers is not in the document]
Lancasterwu: This has been deleted since we are not using sample csv file to test `read_student_file.py`

gatorgrouper/tests/test_read_student_file.py
Michionlion: It could probably be either `generated_csv` or something like that, but your current name is fine with with me.

gatorgrouper/tests/test_num_group.py
Michionlion: @Lancasterwu the docstring fix is the only thing left I think then!

Lancasterwu: Thx, just fixed it!

Lancasterwu:
Hi @Michionlion, I think this PR is ready to be merged.

Issue #161, opened 2019-02-22
Write test cases for Django models and views


aubreypc:
As we begin to expand the functionality of our Django application, we should look into best practices for testing models and views.

PR #193, opened 2019-03-01
Added test cases for Django models

enpuyou:
# Description of Pull Request

This PR adds test cases for Django models. It also creates several new test files for future testing for views, admin, and urls.  In able to use `pytest` and perform a better test, `pytest-django` and `mixer` are added into the pipfile. In addition, the PR adds a path for `DJANGO_SETTINGS_MODULE` in the `pytest.ini`

It fixes the models part of the #161 , since the views are not finished implemented. 

## Type of Change

- [ ] Bug fix
- [ ] Breaking change
- [ ] New feature
- [ ] Documentation update
- [x] Improving Test Coverage

## Tags
@enpuyou


gatorgrouper/tests/test_models.py
```diff
@@ -0,0 +1,95 @@
...
+        """test professor model"""
+        obj = mixer.blend('gatorgrouper.Professor')
+        # it creates a professor instance
+        assert obj.pk == 1
+
+    # pylint: disable=R0201
+    def test_str(self):
+        """test __str__ method """
+        obj = mixer.blend('gatorgrouper.Professor', last_name='K', first_name='Greg')
+        result = obj.__str__()
```
Michionlion: This should be called with the actual `str` function, like: `result = str(obj)`. That is what the magic method `__str__` enables.

gatorgrouper/tests/test_models.py
```diff
@@ -0,0 +1,95 @@
...
+    # pylint: disable=R0201
+    def test_model(self):
+        """test grouped student model"""
+        obj = mixer.blend('gatorgrouper.Grouped_Students')
+        # it creates a grouped students instance
+        assert obj.pk == 1
+
+    # pylint: disable=R0201
+    def test_str(self):
+        """test __str__ method """
```
Michionlion: This does not test the output of `__str__`, just that it is the same as `str`.

gatorgrouper/tests/test_models.py
enpuyou: Fixed in d3fe167742a644c09feb5d6e682a21c453916028

gatorgrouper/tests/test_models.py
enpuyou: Fixed in d3fe167742a644c09feb5d6e682a21c453916028

Lancasterwu:
Hi @enpuyou, I just fixed a merge conflict so it can be tested by Travis CI.

PR #207, opened 2019-03-04
Exception handling for a csv file not containing a header

thomad74:
# Description of Pull Request

The changes in this Pull Request show work towards refactoring the `read_student_file.py` file to specifically provide exception handling for csv files that do not contain a header.

Works on #184 .

## Type of Change

- [x] Bug fix
- [ ] Breaking change
- [ ] New feature
- [ ] Documentation update
- [ ] Other 

## Tags

@Lancasterwu @Michionlion


thomad74:
This change decreases the code coverage in the `read_student_file.py`, but @Lancasterwu and I will look into testing the added lines of code to increase the coverage back to 100%.

Lancasterwu:
capture

While working on writing another test function which tests the csv without a header, I am wondering how did this code passing Travis CI..

Lancasterwu:
If this commit passes Travis CI, it is probably some issue due to using win32 instead of linux,

Lancasterwu:
Based on [the most recent Travis CI log](https://travis-ci.org/GatorEducator/gatorgrouper/builds/501872104?utm_source=github_status&utm_medium=notification), the test coverage did not increase with the right test case, the `has_header` does not provide the correct boolean. The recognizing function needs to be changed.

Lancasterwu:
It was actually complicated to understand [this](http://cat2.mit.edu/dh_upload/lib/miniconda3/pkgs/python-3.6.3-hc9025b9_1/lib/python3.6/csv.py), lol.

gatorgrouper/utils/read_student_file.py
```diff
@@ -13,17 +13,30 @@ def read_student_file(filepath):
 
     # read the raw CSV data
     with open(filepath, "r") as csvfile:
-        csvdata = list(csv.reader(csvfile, delimiter=","))
+        has_header = csv.Sniffer().has_header(csvfile.read(1024))
```
Michionlion: Is the limit `1024` needed?

gatorgrouper/utils/read_student_file.py
```diff
@@ -13,17 +13,30 @@ def read_student_file(filepath):
```
Michionlion: You need to handle the exceptional case.

gatorgrouper/utils/read_student_file.py
thomad74: It runs `has_header` for the first 1024 bytes or whatever the arbitrary number is. If there is no number, it runs it for the entirety of the csv, so it is not completely necessary. The only thing the limitation does is improve efficiency when running the function for csv files.

gatorgrouper/tests/test_read_student_file.py
```diff
@@ -17,12 +18,43 @@ def generate_csv(tmpdir_factory):
...
+    """ Generate a tempory sample csv """
+    fn = tmpdir_factory.mktemp("data").join("csvNg1.csv")
+    df_list = {
+        "NAME": ["delgrecoj", "delgrecoj2"],
+        "Q1": ["True", "True"],
+        "Q2": ["True", "True"],
+        "Q3": ["False", "True"],
+        "Q4": ["True", "True"],
+    }
+    df = pd.DataFrame(df_list)
```
Michionlion: There is no need to use `pandas` here... why not just write the `csv` manually or with a `csv.DictWriter`? No need to add a dependency for something so simple.

Lancasterwu: Hi @Michionlion, this is because `csv.DictWriter` can only accept the parameter of dictionary which means header is required as the key in dictionary. I did a lot of research and I still did not find the way to generate csv without header by using csv package. However, by using `pandas` package, I can set `header=False` when I am implementing `to_csv` function. Which will generate a csv without header.

gatorgrouper/utils/read_student_file.py
```diff
@@ -13,17 +13,30 @@ def read_student_file(filepath):
...
+            responses.append(temp)
+    else:
+        for record in csvdata:
+            temp = list()
+            temp.append(record[0].replace('"', ""))
+            for value in record[1:]:
+                if value == "True":
+                    temp.append(True)
+                elif value == "False":
+                    temp.append(False)
```
aubreypc: To elaborate on my earlier suggestion of adding parsing for floats, one way of accomplishing this would be something along the lines of:
```
else:
    try:
       temp.append(float(value))
    except ValueError: # value cannot be converted to float
        pass
```

gatorgrouper/utils/read_student_file.py
```diff
@@ -13,17 +13,30 @@ def read_student_file(filepath):
 
     # read the raw CSV data
     with open(filepath, "r") as csvfile:
-        csvdata = list(csv.reader(csvfile, delimiter=","))
+        has_header = csv.Sniffer().has_header(csvfile.read(1024))
 
+    with open(filepath, "r") as csvfile:
```
aubreypc: Could you change the argument to be called `csvfile` and check `type(csvfile)` at the start of the function to see if a path string or a file object is being passed in instead of always opening the file from a path string? In the `feature/mvp` branch we ran into a situation where we wanted to call this function but only had access to an object and not a path, and it would probably be better to change it here rather than duplicate code over there.

gatorgrouper/utils/read_student_file.py
aubreypc: Instead of using error handling, this could also be done with a regex string matching only floats (probably less readable imo).

gatorgrouper/tests/test_read_student_file.py
aubreypc: If we just need to generate CSV files for test cases (i.e. as long as we don't have to do anything with properly escaping commas in the strings, etc), then I believe we could make a headerless CSV file by doing
```
data = [
    # optionally include headers as the first entry
    ["delgrecoj", "True", "True", "False", "True"],
    ["delgrecoj2", "True", "True", "True", "True"]
]
csv_string = ""
for entry in data:
     csv_string += ",".join(entry) +"\r\n"

# write csv_string to the output file
...
```

Lancasterwu: I have tried many different ways including this. You can find that in one of the commit history. It turns out that they would not generate the non-header-csv properly. For some reason it will always recognize the first line as the header, and I finally get it work by using `pandas`.

Michionlion: print out the actual string generated by `pandas`, and use that.

gatorgrouper/utils/read_student_file.py
```diff
@@ -26,8 +26,11 @@ def read_csv_data(filepath):
             for value in record[1:]:
                 if value == "True":
                     temp.append(True)
-                elif value == "False":
-                    temp.append(False)
+                else:
+                    try:
+                       temp.append(float(value))
+                    except ValueError: # value cannot be converted to float
+                        pass
```
Michionlion: Does this still handle `False` then?

Lancasterwu: I don't think it does, this PR need to be reverted to the older version.

gatorgrouper/utils/read_student_file.py
Lancasterwu: Hi @aubreypc, I don't think this should be handled in this PR.

gatorgrouper/utils/read_student_file.py
Lancasterwu: This should be handled in other PR, but not this one. The purpose of this PR just just to implement header handling.

gatorgrouper/tests/test_read_student_file.py
Lancasterwu: @Michionlion I will check that

gatorgrouper/utils/read_student_file.py
Michionlion: Why reverted? It can probably be added here...

gatorgrouper/tests/test_read_student_file.py
Lancasterwu: I have tried and it does not work. The only way to generate csv with `tmpdir_factory` besides using `pandas is to use `csv.DictWriter(csvfile, fieldnames=headers)` and in this situation the header will always be recognized by `read_student_file` since `csv` does not have the feature to set `header=False`.

Lancasterwu: Actually, nvm, I think I figured it out. I thought directwrite is the only way to generate csv.

PR #218, opened 2019-03-05
Update to .gitignore to reflect additions

Alex-Yarkosky:
# Description of Pull Request

I have updated the `.gitignore` file to include all relevant material that should be excluded from the repository to the best of my knowledge. The updates were from [gitignore.io](https://gitignore.io/).

Fixes `.gitignore` file

## Type of Change

- [ ] Bug fix
- [ ] Breaking change
- [ ] New feature
- [X] Documentation update
- [ ] Other 

## Tags
 @Alex-Yarkosky


.gitignore
```diff
@@ -5,3 +5,440 @@ __pycache__
...
+# Created by https://www.gitignore.io/api/vim,git,macos,linux,django,python,eclipse,windows,intellij,notepadpp
+# Edit at https://www.gitignore.io/?templates=vim,git,macos,linux,django,python,eclipse,windows,intellij,notepadpp
+
+### Django ###
+*.log
+*.pot
+*.pyc
+__pycache__/
+local_settings.py
+db.sqlite3
```
Michionlion: We should ignore *all* `sqlite3` file!

.gitignore
```diff
@@ -5,3 +5,440 @@ __pycache__
 gatorgrouper_django/gatorgrouper/migrations/
 /.coverage
```
Michionlion: This line is already taken care of in the below gitignore.

.gitignore
```diff
@@ -5,3 +5,440 @@ __pycache__
...
+
+# Django stuff:
+
+# Flask stuff:
+
+# Scrapy stuff:
+
+# Sphinx documentation
+
+# PyBuilder
```
Michionlion: There's some extra lines here that are seemingly empty.

.gitignore
```diff
@@ -5,3 +5,440 @@ __pycache__
...
+# Rope project settings
+
+# mkdocs documentation
+
+# mypy
+
+# Pyre type checker
+
+### Python Patch ###
+.venv/
```
Michionlion: This is already ignored.

.gitignore
sutterj: @Michionlion 
I want to suggest that we actually leave the db in the repo, but if you think otherwise then there are a couple of things that should end up getting documented somewhere.  You cannot run django normally if you do not have the db without first running the commands `python3 manage.py makemigrations` and `python3 manage.py migrate` to create it and then running `python3 manage.py createsuperuser` to be able to access the admin pages.  All of the models and logins get added to the database and if you don't have it in the repo then they must get recreated as part of the setup.

When this eventually goes onto AWS, the database does need uploaded there as the site will not correctly function without it.  There is a way to have it so that when you push to master the site will automatically update, but this will not work correctly unless the db is committed.

The only reason that I can think of not to commit the database is for security reasons, but the development version of the tool shouldn't have any sensitive information.

Michionlion: Okay, that makes sense. I have seen merge conflicts with the database as well however, which may be complicated for students to resolve, since they will need to specify `--ours` or `--theirs` for the database file.

Michionlion: @Alex-Yarkosky Can you remove the `sqlite3` ignore?

Alex-Yarkosky:
@Michionlion @sutterj I believe I have made all of the requested revisions, feel free to let me know if there is anything that I have missed or anything else that should be changed. Thanks!

Michionlion:
> @Michionlion @sutterj I believe I have made all of the requested revisions, feel free to let me know if there is anything that I have missed or anything else that should be changed. Thanks!

Yep, looks good! You just need to resolve the conflicts and get one more review.

Issue #112, opened 2019-01-29
Perform Robustness Testing with Hypothesis


gkapfham:
While GatorGrouper currently has a high coverage test suite, it is good idea for us to perform additional testing, perhaps by using a tool like Hypothesis to generate random input values. These tests could be marked separately and, perhaps, not during during the Travis CI build. However, this new type of testing could be useful for finding hard-to-discover bugs in "corner cases" of, for instance, parameter verification.

libailong:
https://hypothesis.readthedocs.io/en/latest/
This website helps a lot to get a general understanding of Hypothesis.
https://hypothesis.readthedocs.io/en/latest/data.html
This tells us what we can generate from the Hypothesis. I am working on the test for text_display.py and try to figure out what type of input is for display and how to generate that? Do you guys have any ideas?

mmarconi:
https://scipy-lectures.org/packages/statistics/index.html#id13
Here is another link with some info about Hypothesis testing in Python

huangs1:
This should not have been closed since we still need to create more hypothesis tests to be more confident in the robustness for more functions. My team is currently working on creating more test cases using hypothesis, there should be a PR that will fix this issue soon.

PR #220, opened 2019-03-05
added hypothesis test case

mmarconi:
# Description of Pull Request

Added in a hypothesis test case.
The test case I wrote uses hypothesis to test the second function in test_group_method.py.

Works #112 

## Added in a test case

- [ ] Bug fix
- [ ] Breaking change
- [ ] New feature
- [ ] Documentation update
- [x] Other Code Coverage

## Tags

@huangs1 @Michionlion @Lancasterwu 

@libailong


gatorgrouper/tests/test_group_method.py
```diff
@@ -1,7 +1,18 @@
...
+from hypothesis import Verbosity
+from hypothesis.strategies import integers
+
+# from hypothesis.strategies import text
+import pytest
 from utils import group_random
 from utils import group_rrobin
 
+# import parse_arguments
+# import gatorgrouper
```
Michionlion: There are a number of unneeded comments in this file; they should be removed if they are not needed.

gatorgrouper/tests/test_group_method.py
```diff
@@ -60,6 +71,25 @@ def test_group_random_extra():
     assert num_group == 2
 
 
+@given(grpsize=integers(min_value=1, max_value=3))
+@settings(verbosity=Verbosity.verbose)
+@pytest.mark.hypothesisworks
+def test_group_random2(grpsize):
+    """This hypothesis test will test the test_group_random_extra"""
```
Michionlion: This docstring is incorrect -- this test only tests the `group_random_group_size` method. It is, however, a good model for a hypothesis test.

gatorgrouper/tests/test_group_method.py
```diff
@@ -60,6 +71,25 @@ def test_group_random_extra():
...
+        ["Marvin", False, False, True, True],
+        ["Evin", True, True, True, False],
+        ["Nikki", True, True, False, False],
+        ["Nick", True, False, True, False],
+        ["Dan", False, True, False, True],
+    ]
+    returned_groups = group_random.group_random_group_size(responses, grpsize)
+    size_count = grpsize
+    assert len(returned_groups[0]) == size_count
+    assert grpsize == size_count
```
Michionlion: This second assertion means nothing -- `size_count` was just set to be `grpsize`. You can completely remove the mention of `size_count`. You should also decide if you need to check for more than just the size of the first group (`returned_groups[0]`). Do all groups have to have the same size to pass this test? What is the intended behavior for `group_random_group_size`?

gatorgrouper/tests/test_group_method.py
mmarconi: I changed the docstring.

gatorgrouper/tests/test_group_method.py
mmarconi: I deleted the second assertion. I will consult with my team about about checking the size of more than just the first group.

Issue #206, opened 2019-03-04
Refactoring group_scoring.py and Associated Functions


ALEXANDERB82:
# Description of the Issue
The file group_scoring.py has functions in it that are named in a very confusing way. Also the docstrings that are present do not describe exactly what the function are doing. group_scoring.py needs to be refactored.

## Add Detailed Comment with More Information
The name of the function score_groups found withing group_scoring.py implies that the function is scoring multiple groups which in fact it does not. This can make it hard for people to understand the purpose of the function. The docstring that is provided for the function states that the function scores multiple groups however the function does not do that. The function score_groups creates an array of the scores of all of the groups and return it as well as return the average of all of the scores for the groups. The function score_groups needs to be refactored so that there is a greater generalization so that other functions that use it are less dependent.

## Steps to Reproduce Issue
In order to recreate this issue you would need to produce the code below in the group_scoring.py file.
```
""" score groups by approximate level of diversity """


def score_group(student_identifers):
    """ score single group """
    score = 0
    for student in student_identifers:
        for category in range(len(student[1:])):
            if student[category + 1] is True:
                score += 1
    return score


def score_groups(student_groups):
    """ score multiple groups """
    scores = []
    ave = 0
    for group in student_groups:
        scores.append(score_group(group))
        ave += score_group(group)
    ave = int(ave / len(student_groups))
    # returns the scores of each group as well as the average score
    return scores, ave
```
### Assigned Developers
@ALEXANDERB82

ALEXANDERB82:
This issue has been addressed in the branch; [ enhance/debug-improve-group_scoring](https://github.com/GatorEducator/gatorgrouper/tree/enhance/debug-improve-group_scoring).

PR #222, opened 2019-03-05
Refactor/improve generality group creation

### LINKED ISSUE #217: Increasing Generality to Decrease the Number of Dependent Modules for group_scoring.py
# Description of the Issue
All of the different group creation algorithms are dependent on group_scoring.py but are all located within different files. The code needs to be refactored and re designed so that there is more generality overall.

Currently all of the group creation algorithms are located in various different files, however they should all be placed within one file. Each group creation algorithm should simply be a function within a group_creation.py file that would be responsible for each of the algorithms. The code should also then be changed so that whenever the old files for group creation were referenced within the project it is replaced with a reference to group_creation.py instead. 

## Steps to Reproduce Issue
* Create separate files for each group creation algorithm

### Assigned Developers
@ALEXANDERB82 
@Lancasterwu

ALEXANDERB82:
# Description of Pull Request

This pull request creates a file named group_creation.py that contains all of the grouping algorithms. It also removes the old files that contained the grouping algorithms separately; group_rrobin.py, and group_random.py. 

This Pull request also changes the name of the function score_groups in group_scoring.py to calculate_avg. And adds documentation more fitting to the purpose of the function.

Lastly this Pull Request refactors the parse_arguments.py file by getting rid off the method flags and using --method instead.

Fixes #217 and #206 

## Type of Change

- [ ] Bug fix
- [ ] Breaking change
- [x] New feature
- [x] Documentation update
- [x] Other (Refactoring)

## Tags

@Lancasterwu @gkapfham @Michionlion @toccinAC @aubreypc


gatorgrouper/tests/test_arguments.py
```diff
@@ -13,7 +13,9 @@ def test_parse_arguments1():
     assert parsed_args.logging_level == logging.ERROR
     assert parsed_args.group_size == defaults.DEFAULT_GRPSIZE
     assert parsed_args.students_file == defaults.DEFAULT_CSVFILE
-    assert (parsed_args.grouping_method == group_random) is False
+    # The following line may need to be changed
+    # Unsure of the group_creation references
+    assert (parsed_args.grouping_method == group_creation) is False
```
Michionlion: This does not make sense now and didn't before -- `grouping_method` is a string, and should probably reference a constant in `constants.py`.

gatorgrouper/utils/group_scoring.py
```diff
@@ -11,8 +11,8 @@ def score_group(student_identifers):
     return score
 
 
-def score_groups(student_groups):
-    """ score multiple groups """
+def calculate_ave(student_groups):
```
Michionlion: This should probably use the abbreviation `avg` instead of `ave`.

gatorgrouper/utils/group_creation.py
```diff
@@ -0,0 +1,164 @@
+"""Contains all of the gorup creation algorithms"""
```
Michionlion: I would suggest adding a single interface method to this module that takes two string constants from `constants.py` (one for grouping method and the other for whether to use the given number for group size or group number) and then calls the correct grouping function based on those arguments. This will make all API and testing calls the same function, enabling better, easier to read code.

gatorgrouper/utils/group_scoring.py
ALEXANDERB82: I have changed the abbreviation from ave to avg in all necessary locations.

gatorgrouper/utils/group_creation.py
aubreypc: I agree that there should be a single interface for all the grouping functions. It may be best to use keyword arguments rather than positional arguments so that parameters can be optional and default values can be provided. 

Maybe the function could determine which grouping algorithm should be used (if one is not explicitly specified) based on which keyword arguments are provided? For example, if a "compatibility" parameter is passed in, it would decide to use the graph partitioning algorithm since the others do not take compatibility into account.

Any thoughts on this idea @ALEXANDERB82 @Michionlion ?

gatorgrouper/tests/test_arguments.py
aubreypc: In addition to @Michionlion's comment, a suggestion about the syntax: `var1 != var2 ` is generally easier to read than `(var1 == var2) is False`

gatorgrouper/utils/group_creation.py
Michionlion: Sounds good, some requirements should be defined and then this can be designed!

everitt-andrew:
Just wanted to mention that with the updates to the grouping methods and structure, a lot of the tests were broken and the code coverage currently sits at 84%

PR #229, opened 2019-03-06
Add a graph partitioning algorithm for grouping

### LINKED ISSUE #190: Represent grouping as a Graph Partition Problem and implement a Kernighan-Lin heuristic
## Background Info

In the **graph partition problem**, we must partition the vertex set _V_ of a graph into _k_ subsets such that that the edge-cut (total weight of all edges which go between different subsets) is minimized. In our case, each vertex in the graph would represent a particular student, and for any two vertices _u_ and _v_ we create an edge _e_ between them if and only if those two students are allowed to be in the same group. Then we assign a weight to _e_ representing the compatibility of the two students. With this setup, the edge-cut represents the total "unused compatibility" of the grouping, i.e. the sum of all compatibility scores for students who _aren't_ in the same group -- we want to minimize this.

While graph partitioning is NP-hard, it has some fairly fast heuristics. Kernighan-Lin is one such option which applies in the case of bisecting the graph into 2 groups (_k_ = 2). By recursively applying Kernighan-Lin, we can partition into _k_ = 2^_n_ groups for any _n_ >= 1 in roughly O(log2(_k_) * |_V_|^2) time. However, other heuristics are available which are similar to Kernighan-Lin while generalizing to a _k_-way partition rather than a bisection, and perhaps without the logarithmic increase in complexity.

- Brief videos explaining graph partition problem and Kernighan-Lin: [1](https://www.youtube.com/watch?v=MMlf66PQdN8), [2](https://www.youtube.com/watch?v=2swNkneUmg0) 
- [Documentation for the Kernighan-Lin implementation in the NetworkX library](https://networkx.github.io/documentation/latest/reference/algorithms/generated/networkx.algorithms.community.kernighan_lin.kernighan_lin_bisection.html)
- [A paper describing how to optimize for arbitrarily many objectives at once](http://glaros.dtc.umn.edu/gkhome/fetch/papers/mo.pdf) -- this will help us assign edge weights

## Prototype implementation

We should begin by getting recursive K-L on one objective working before moving on to multi-objective optimization. This will involve:

- Computing compatibility between students
- Constructing a graph representing student conflicts (student pair conflict => no edge between vertex pair) and assigning compatibility scores as edge weights
- Running K-L recursively to partition students into 2^_n_ groups

aubreypc:
# Description of Pull Request

This PR implements a new grouping algorithm based on the Kerighan-Lin graph partitioning heuristic. This new algorithm is capable of handling conflicts between students, as well as assigning compatibility scores to pairs of students. Essentially, the algorithm seeks to minimize the number of "missed connections" between compatible students by minimizing the total sum of all compatibility scores between students in different groups.

Fixes #190 

## Type of Change

- [ ] Bug fix
- [ ] Breaking change
- [x] New feature
- [ ] Documentation update
- [ ] Other 

## Tags

@aubreypc


gatorgrouper/utils/group_graph.py
```diff
@@ -0,0 +1,97 @@
...
+    print("Partition:")
+    for p in partition:
+        groups.append([responses[i] for i in p])
+        print([G.nodes[i]["weight"] for i in p])
+    print("Total cut size:", total_cut_size(G, partition))
+    return groups
+
+
+if __name__ == "__main__":
+    students = ["one", "two", "three", "four", "five", "six", "seven", "eight"]
```
Michionlion: This needs to be a test case

Lancasterwu:
Hi @aubreypc , I just went through this with @shafferz. I will implement this algorithm with CLI of gatorgrouper, and my team will create test case for it.

Michionlion:
> Hi @aubreypc , I just went through this with @shafferz. I will implement this algorithm with CLI of gatorgrouper, and my team will create test case for it.

Should we merge this to master then, or will you work in `feature/graph-algorithms`?

Lancasterwu:
> > Hi @aubreypc , I just went through this with @shafferz. I will implement this algorithm with CLI of gatorgrouper, and my team will create test case for it.
> 
> Should we merge this to master then, or will you work in `feature/graph-algorithms`?

I think I will work on this with a new branch after it is merged in to the master. However,  I am wondering why `codecov` passed with no test case to this function.

shafferz:
Nice implementation, @aubreypc! The math of this is a little mind-boggling, but I think I'm starting to figure it out.

As @Lancasterwu said, I worked with him through this algorithm. I know you addressed this in Issue #190, but I am still not sure of the way that the student compatibility scores are working here (or how the professor would go about setting these up). 

Does the professor need to set a compatibility score for each individual (unique) pair of students as input? If this is the case, wouldn't the professor then need to provide _n choose 2_ inputs for n students? If that is the case, for a class size comparable to ours (lets say, 25 students), then there'd be upwards of 300 values that the professor would have to input manually.

I noticed that there was a list `preferences` that defaults to the value `None`, which may be related to this problem. Would this `preferences` list be able to be used to modify specific weights between students to be lower (or zero)? If not, is there _any_ way to reduce a weight between two students to zero, for the purpose of conflict avoidance?

I also noticed that the numbers of groups can only be powers of 2. Is this something that can be solved or updated? Do you foresee this being a problem?

I think the answers to these questions would be helpful to my team (especially @ohnoanarrow and myself) for the conflict handling that we are working with. 

Thanks, and sorry for the long read!

aubreypc:
> Nice implementation, @aubreypc! The math of this is a little mind-boggling, but I think I'm starting to figure it out.
> 
> As @Lancasterwu said, I worked with him through this algorithm. I know you addressed this in Issue #190, but I am still not sure of the way that the student compatibility scores are working here (or how the professor would go about setting these up).
> 
> Does the professor need to set a compatibility score for each individual (unique) pair of students as input? If this is the case, wouldn't the professor then need to provide _n choose 2_ inputs for n students? If that is the case, for a class size comparable to ours (lets say, 25 students), then there'd be upwards of 300 values that the professor would have to input manually.
> 
> I noticed that there was a list `preferences` that defaults to the value `None`, which may be related to this problem. Would this `preferences` list be able to be used to modify specific weights between students to be lower (or zero)? If not, is there _any_ way to reduce a weight between two students to zero, for the purpose of conflict avoidance?
> 
> I also noticed that the numbers of groups can only be powers of 2. Is this something that can be solved or updated? Do you foresee this being a problem?
> 
> I think the answers to these questions would be helpful to my team (especially @ohnoanarrow and myself) for the conflict handling that we are working with.
> 
> Thanks, and sorry for the long read!

Hi @shafferz,

Thanks for the feedback! Your concern about compatibility data entry is very valid. The original plan was to assign a weight to each vertex/student based on the CSV data (with floats rather than just True/False), and then use this to automatically determine the compatibility of each pair of students. The existing `compatibility` function in this PR probably does not do this in the best way, and so I've started thinking that we should just go with our original plan of allowing students to self-report, except report their _preferences_ instead of skills. (See #235 and feature/survey) This way the effort of collecting the data will be minimized, since each student will only have a couple preferences, and since this allows entry of preferences to be optional.

Re: the existing `preferences` parameter, this is admittedly a very bad name for it -- it's terminology borrowed from [this paper](http://glaros.dtc.umn.edu/gkhome/node/93). It is a "preferences vector" which give scalars to the different float-containing columns in our CSV, so that we can make a normalized sum and reduce our node weight tuple into a single number for node weight, thereby allowing us to optimize for it. So if we are able to define a good compatibility function, this could allow the professor to have finer control over which "objectives" are more important. That said, it's worth just removing this to avoid confusion if we're taking it in a different direction.

I'm not entirely sure about conflict avoidance -- I was thinking originally that if we just don't `G.add_edge(...)` when there's a conflict between a certain pair, then they wouldn't end up in the same group -- but since there's still connectivity between the conflicting pair and other mutual teammates, it seems like it would still be possible for them to end up in the same group. Testing would be needed on this. Not sure exactly how it would play out with zero / negative edge weights.

As for the powers of 2, that's unfortunately a limitation of Kernighan-Lin -- it can only bisect, and so `recursive_kl` can only recursively bisect. This is definitely dangerous in terms of leaving room for user error. We will have to make sure to validate the data we receive from the user, and probably design our grouping UI in a way which avoids invalid inputs. There are also more general algorithms in the spirit of Kernighan-Lin, but there's probably not enough time to implement them from scratch.

Michionlion:
I would highly suggest using skills instead of preferences to generate compatibility scores, but the design is ultimately yours. Just '@' me when this should be reviewed and merged.

aubreypc:
I think a good compromise would be to have the compatibility score be determined by the student CSV data, and then have preferences act as a scalar multiple to determine the compatibility score.

@Lancasterwu Is your team still planning to write test cases for the algorithms?

Michionlion:
@aubreypc I'm extremely hesitant to say that getting data from students is a good idea. I would highly suggest, at the very least, to just require one numerical representation of "skill" for each student, and then generate groups where that level is averaged across the class. Conflicts can be incorporated into that.

Alex-Yarkosky:
I still think that it would best serve the professor to by default assume that all students work well together with each other, represented by the highest compatability score in my mind, and then the professor can change compatability between students to a lower value in they know that they either do not work well together or have some form of conflict. These would essentially be exceptions to the logic that each and every student can work fine with any and all of the other students.

aubreypc:
@Michionlion To clarify, the only data collected from students (at least in my own mental model of the product) is a list of other students who have been marked as preferred, and only if the professor configures the assignment to take preferences into account. When I say "student CSV data" I mean the columns in the CSV file provided by the professor.

My intuition is that this graph partitioning algorithm would be suited for the class-wide averaging which you suggest, but further experimentation is required there. I would recommend opening an issue for your suggestion, maybe @Alex-Yarkosky's team could work on that.

PR #232, opened 2019-03-06
Refactored parse_arguments.py, removed bugs and added test cases

### LINKED ISSUE #194: The `parse_argument` Should not Handle Group Number 0 and 1
Based on Issue https://github.com/GatorEducator/gatorgrouper/issues/184, I found that:

The function  `parse_argument.py` has a bug in `check_valid_num_group` which allows the program to handle input `--num-group 0` and `--num-group 1`, It should not be able to. 

```
    if numgrp > students_list_length:
        logging.error("Number of groups: %d", numgrp)
        logging.error("Number of students: %d", students_list_length)
```

This is a bug because spliting the class to 0 group or 1 group does not make any sense. Because of this, ` test_check_number_groups_size_zero` and `test_check_number_groups_size_one` in `test_num_group.py` are currently passing, which they shouldn't.

@reibelj will make a PR to fix this bug and the related test functions.

### LINKED ISSUE #202: Refactoring parse_arguments.py to remove exit statements
# Description of the Issue
Based on [Issue#184](https://github.com/GatorEducator/gatorgrouper/issues/184)
There are 2 exit statements in parse_arguments.py that are difficult to test. Refactoring may ease the testing process and increase the efficiency.

Currently there are 2 exit statements on lines 93, and 104. These statements are difficult to test and could be more efficient. Slightly refactoring the program will allow to remove these exit statements and replace them with return statements. This will remove the requirement to test the statements and increase code coverage and efficiency of parse_args.py

### Assigned Developers
@Lancasterwu  @reibelj

baldeosinghm:
# Refactored parse_arguments.py, removed bugs and added test cases

This pull request contains refactored code in `parse_arguments.py`.  The quit() commands were removed to make testing of the program easier and also make the code more efficient.  In its place, the string "Invalid entry" was coded.  Furthermore, I made test cases to increase code coverage with `test_arguments.py` (@enpuyou also helped out).  Also, there was a bug that was fixed in `parse_arguments)` where student groups of 1 and 0 were being considered, when they shouldn't.

Fixes #202, fixes #194 

## Type of Change

- [x] Bug fix
- [ ] Breaking change
- [x] New feature
- [ ] Documentation update
- [x] Testing: Test cases were made in `test_arguments.py`

## Tags

@enpuyou - helped in creating a test case for `parse_arguments.py`


gatorgrouper/utils/constants.py
```diff
@@ -1,3 +1,12 @@
 """ These are global constants """
 
+
 ALGORITHM_ROUND_ROBIN = "rrobin"
+ALGORITHM_RANDOM = "random"
+DEFAULT_METHOD = "random"
+DEFAULT_CSVFILE = "/home/w/wuj/cs203S2019/lab/students.csv"
```
Michionlion: This should be removed or changed.

gatorgrouper/utils/gatorgrouper.py
```diff
@@ -14,8 +14,12 @@
 
 if __name__ == "__main__":  # pragma: no cover
 
-    # parse the arguments and display welcome message
+    # parse the arguments
     GG_ARGUMENTS = parse_arguments.parse_arguments(sys.argv[1:])
+    if (GG_ARGUMENTS == "Invalid entry"):
```
Michionlion: This is not checking for the right string -- the string returned should be declared in `constants.py`.

tests/conftest.py
```diff
@@ -0,0 +1,49 @@
...
+    fn = tmpdir_factory.mktemp("data").join("csvNg1.csv")
+    df_list = {
+        "NAME": ["delgrecoj", "delgrecoj2"],
+        "Q1": ["True", "True"],
+        "Q2": ["True", "True"],
+        "Q3": ["False", "True"],
+        "Q4": ["True", "True"],
+    }
+    df = pd.DataFrame(df_list)
+    df.to_csv(str(fn), index=False, header=False)
```
Michionlion: You should write this csv manually (by just writing text to a file if need be), instead of using `pandas`, which is unneeded dependency bloat.

tests/test_read_student_file.py
```diff
@@ -0,0 +1,12 @@
...
+# pylint: disable=redefined-outer-name
+import pytest
+
+from gatorgrouper.utils import read_student_file
+
+
+def test_read_student_file(generatcsv):
+    """ Test read_student_file """
+    expectedoutput = [["delgrecoj", True, True, False, True]]
+    assert read_student_file.read_student_file() == expectedoutput
```
Michionlion: How does this pass? `read_student_file` expects a filepath to be given as a parameter -- that is what `generatcsv` is for (also a misspelling in that).

the_main.py
```diff
@@ -0,0 +1,85 @@
+""" GatorGrouper randomly assigns a list of students to groups """
```
Michionlion: This file should not be named `the_main.py`, instead use the either `main`, or the name of the program. Since this is the entrypoint for the program, I would suggest `gatorgrouper.py`, or at least `gatorgrouper_cli.py`.

the_main.py
```diff
@@ -0,0 +1,85 @@
...
+        )
+        logging.info("GatorGrouper randomly ordered the students:")
+        logging.info(
+            "\n %s",
+            display.create_escaped_string_from_list(SHUFFLED_STUDENT_IDENTIFIERS),
+        )
+
+        # generate the groups and display them
+        # pylint: disable=bad-continuation
+        if (
```
Michionlion: This kind of logic should not be in the entry point program -- instead, an interface or API method needs to be created that can take, for instance, a grouping method and other arguments, and accomplish this task. That way, the logic can be tested easily.

Lancasterwu:
This branch should be reverted to the version before merged from mvp-terminal-only. Hence I will close it for now.

PR #240, opened 2019-03-09
Implement Graph Algorithm Grouping Method with CLI

### LINKED ISSUE #190: Represent grouping as a Graph Partition Problem and implement a Kernighan-Lin heuristic

### LINKED ISSUE #234: Improve generality of read_students_file
# Description of the Issue

Reading student data from a CSV file occurs in both the Django app and the command-line app, but the existing read_students_file function only applies to the command-line context. 

1. Currently the function takes a file path string as an argument, but in our Django app we have a case in which we need to pass in a file-like object instead of a path. To avoid duplicating code, this function should be refactored to accept both types, skipping the process of opening the file if the argument is not a string.
2. Rather than parsing just true and false values in the CSV, integers and floats should also be parsed.

Lancasterwu:
# Description of Pull Request

Fixes #190 
Fixes #234 

## Type of Change

This PR is the second step of fixing issue #190, after [PR#229](https://github.com/GatorEducator/gatorgrouper/pull/229) and before the testing and documentation.
The following changes have been made:
* Add code to `parse_argument.py` in order to recognize new command: `--graph`, which runs the graph algorithm grouping.
* Add code to `gatorgrouper_cli.py` to allow run the command and get the output of graph algorithm grouping.
* Add code to `read_student_file.py` to make it be able to accept float input and generate it to the list of list which contains float. Fixes issue 234
* Add two new pytest fixtures in `conftest.py` which allow developer to write test case with temporary csv file that contains float. One is the csv that has header, another one is the csv that does not have header.
* Add two new test function for `read_student_file.py` to test if the float input can be handled properly.
* Restructured `group_graph.py` in order to use the matching input that other files passing in, and the matching output that can be passed to other file.

Note that implementing Graph Algorithm Grouping Method with CLI is not completed, the test case for  `group_graph.py` and the documentation are still missing. In order to match the output and input there are some really bad designs in `group_graph.py` which also need to be refactored.

- [ ] Bug fix
- [x] Breaking change
- [x] New feature
- [ ] Documentation update
- [ ] Other 

## Tags

@GatorEducator/team-lancaster should be responsible to provide documentation and test case for this ASAP. Please contact me @aubreypc @paladp @huangs1 if you find any bug.


gatorgrouper/utils/group_graph.py
```diff
@@ -68,30 +69,34 @@ def group_graph_partition(
...
         groups.append([responses[i] for i in p])
-        print([G.nodes[i]["weight"] for i in p])
-    print("Total cut size:", total_cut_size(G, partition))
-    return groups
+    output = list()
+    # Generate a list of list of list in order to fit the required display input
+    for student in groups:
+        temp = list()
+        for value in student:
+            value += " fakestring"
```
Michionlion: `" fakestring"` looks like a debugging artefact. What is its purpose? If a simple string is needed to hold a place, you should use an empty string, or `None`.

gatorgrouper_cli.py
```diff
@@ -61,6 +62,13 @@
             GROUPED_STUDENT_IDENTIFIERS = group_rrobin.group_rrobin_num_group(
                 SHUFFLED_STUDENT_IDENTIFIERS, GG_ARGUMENTS.num_group
             )
+        elif (
+            GG_ARGUMENTS.grouping_method == constants.ALGORITHM_GRAPH
+            and GG_ARGUMENTS.num_group is not constants.DEFAULT_NUMGRP
```
Michionlion: This if statement would pass if `GG_ARGUMENTS.num_group` was not specified, and therefore `None`. This is probably not the behavior you want.

tests/conftest.py
```diff
@@ -93,6 +93,37 @@ def generate_csv_no_header(tmpdir_factory):
     return str(fn)
 
 
+@pytest.fixture(scope="session")
+def generate_csv_float(tmpdir_factory):
```
Michionlion: This is my own opinion, but you should use the same method you used for `generate_csv_float_no_header` here, instead of the `csv` import.

aubreypc:
Refactored `compatibility()`:
- Supports multiple measures of compatibility
- Supports passing in custom measures of compatibility
- Supports using different measures for different columns of the data
- Renamed `preferences` -> `objective_weights` to avoid confusion

gatorgrouper/utils/group_graph.py
Lancasterwu: As I mentioned in issue #241, this code is due to that I need to make the output of this function to be the list of the list of the list. I think there should be a way to do this better than what I did, and that should be handled by the next pull request.

gatorgrouper_cli.py
Lancasterwu: If `GG_ARGUMENTS.num_group` was not specified, wouldn't it use the default num group?

tests/conftest.py
Lancasterwu: Since `def generate_csv` also used csv as import, I would change both of them in the same time which probably should not be under this PR. Also, I'm not sure how to generate a csv which has header which out use `csv`. Should I just make first line different?

gatorgrouper/utils/group_graph.py
aubreypc: @Lancasterwu I believe that if you change `responses[i]` to `inputlist[i]` on line 118 and then `return groups` it should do what you want to it

Lancasterwu: The `inputlist[i]` is a list which contains the student's name at position 0, and there are many weights numbers follow that name. The partition is assigned to the students' name, If we put weights also to those position based on partition, wouldn't that make the output that we don't want? I may have some misunderstanding of the code here.

aubreypc: My understanding was that `inputlist` is a list containing each parsed row of the CSV data (where [1:] become the weights), and that you wanted the final output to be a list of lists of lists, with the the form
```
[
# list of groups
    [
    # first group
        [
        # CSV row corresponding to first student in group
            "student name", True, 0.5, "some other field in data", False, ...
        ]
    ]
]
```
If that's the right format, then my above suggestion should apply, since we need to use the index contained in the partition in order to lookup the entire CSV row rather than the student name in the first position of the row.

Lancasterwu: This is what I thought the structure of the `inputlist` should be. Then I think I got confused about the `partition` part. I have tried and it works. Thanks for the suggestion!

Lancasterwu:
I think this PR should be merged then we can work on the test case, documentation and refactoring in a new branch.

aubreypc:
New feature added to optionally use student preferences as a scalar multiple on the compatibility scores. A command line interface for preferences can be implemented either in this PR or a new one, depending on how soon we want to merge this.

gatorgrouper_cli.py
Michionlion: You're right, my bad!

tests/conftest.py
Michionlion: Yes, you can do the same as your other example -- there's even a comment note on how to generate a header; it's just the first line.

PR #246, opened 2019-03-11
Refactor/change rrobin and gatorgrouper_cli

paladp:
# Description of Pull Request
The major changes this pull request makes are:
- Refactors group_rrobin to address bugs and re-implement the algorithm
- Refactors gatorgrouper_cli to address bugs
- Refactors tests and parse_arguments where necessary to account for previous changes

## Refactoring the Round Robin Grouping Algorithm
- Addressed a bug that caused the distribution of students to groups to behave unexpectedly. Caused by the following code in the master branch: 
``` 
        for response in responses:
            if response[col] is True:
                groups[target_group.__next__()].append(response)
                responses.remove(response)
```
removing the response using `.remove()` caused the for loop to skip certain iterations, leading to inconsistent and possibly wrong behavior where a student is skipped. This had the potential to lead to unbalanced teams. This refactor changes the code to remove student responses after the loop.

- Changed the design of the algorithm to take into account a single priority column and then randomly distribute the rest of the students. The previous algorithm would iterate through every column:

``` 
    for col in columns:
        for response in responses:
            if response[col] is True:
                groups[target_group.__next__()].append(response)
                responses.remove(response)
```
While this sounds good in theory, it has the potential to unfairly distribute students within the groups. For example, consider a case where we have five groups and the first `col` in `columns` is 2 and we have successfully placed a student in each group whose value in column 2 is true. Then we move on to the next `col` in `columns` which is 1. If there are only 3 students whose value in column 1 is true, then they will all be placed in the first three groups, with a zero percent chance of being placed in the other groups. If instead we grouped them randomly, there is equal chance that all five groups will get a student whose value in column 1 is true. While there is an argument that this does not matter, depending on the input repeating this round robin process for every single column can lead to more uneven distribution of students because of this behavior. In this refactor the algorithm no longer iterates through the columns, but instead chooses a random column to do the round robin sorting and then distributes the rest of the students one by one throughout the groups. Some final points:

- Currently the priority column is chosen at random. This is probably not ideal for the final product but I decided to leave it to see what others think, and also because it would likely require changes to the CLI and other parts of the program that I am not comfortable making.
- Currently it is possible that the amount of students with the appropriate column value is less than the number of groups. I would like feedback on how to address this. We could address it like I did in this refactor and distribute them randomly but that would essentially be sorting by random. We could also continue with the round robin anyway. Lastly, we could also ask the user what they would prefer at that point, but that might be complicated.

- Miscellaneous Changes:
This refactor also changes various logging statements to get rid of errors that would appear if the program was ran using the `-v` flag. I'm not entirely sure if the changes are considered best practice, so feedback on that would be nice as well

## Refactoring gatorgrouper_cli 
The refactor to gatorgrouper_cli addresses a logical error in the way that the if statements that handle grouping algorithm method calls were written. For example: 
```
if (
            GG_ARGUMENTS.grouping_method == constants.ALGORITHM_ROUND_ROBIN
            and GG_ARGUMENTS.num_group is constants.DEFAULT_NUMGRP
):
            GROUPED_STUDENT_IDENTIFIERS = group_rrobin.group_rrobin_group_size(
                SHUFFLED_STUDENT_IDENTIFIERS, GG_ARGUMENTS.group_size
)
```
This checks if the user wants to sort by round robin, and if the num_group variable is equal to its default value. If that's true, then we sort using round robin and the group size. This means that if the 
`--num-group` flag is used to explicitly set the value to 3, GatorGrouper will still run using the group size and the output will be incorrect. @aubreypc was able to replicate this bug, and we decided that a change in the logic was necessary. What we actually care about is if the user explicitly set `num_group` or `group_size` **not** if those values are equal to their default. 

By removing default values for num_group and group_size, the refactor changes the comparison above to look like: 
```
 if (
            GG_ARGUMENTS.grouping_method == constants.ALGORITHM_ROUND_ROBIN
            and GG_ARGUMENTS.num_group is None
            and GG_ARGUMENTS.group_size is not None
        ):
```
Now if the user does not enter an explicit value, the variable will simply be `None`. This allows us to check whether the user explicitly entered values and the correct methods are now called. The other if statements have been changed as well, and some new ones have been created to account for new situations. Removing the default values is **likely not best practice** and I am open to hearing about other ways that we could implement the same logic, @aubreypc and I were simply eager to get this change made quickly with the least amount of interference with other parts of the program.

## Other Small Changes and Final Thoughts
There were a variety of other changes made in response to the refactors above, and I would be happy to discuss them, especially because some of the refactors may have had impacts that I missed. Additionally, as discussed with @Lancasterwu we think that round robin should shuffle the students itself, instead of relying on the input to be shuffled beforehand. I did not make this change in this refactor since I was unsure of the best way to accomplish this considering that the code used to shuffle students is in group_random. Thanks to @Lancasterwu for letting me know that round robin needed to be looked at, and to @aubreypc and @everitt-andrew for help with debugging.

Fixes: No issue numbers unfortunately, most of the work done was related to unknown unknowns.

## Type of Change

- [x] Bug fix
- [x] Breaking change
- [ ] New feature
- [ ] Documentation update
- [ ] Other 

## Tags
@Lancasterwu @Michionlion @aubreypc @huangs1 @gkapfham


gatorgrouper/utils/parse_arguments.py
```diff
@@ -33,19 +33,11 @@ def parse_arguments(args):
...
+        "--group-size", help="Number of students in a group", type=int, required=False
     )
 
     gg_parser.add_argument(
-        "--num-group",
-        help="Number of groups",
-        type=int,
-        default=constants.DEFAULT_NUMGRP,
-        required=False,
+        "--num-group", help="Number of groups", type=int, required=False
```
Lancasterwu: Is that necessary to remove the default `numgroup` and `groupsize`? I thought those are helpful though

gatorgrouper/utils/parse_arguments.py
```diff
@@ -98,9 +90,16 @@ def check_valid(args, students_list):
...
+        verified_arguments = True
+    if args.num_group is not None and args.num_group <= students_list_length / 2:
         verified_arguments = True
-    if args.num_group > 1 and args.group_size <= students_list_length / 2:
+    # pylint: disable=bad-continuation
+    if (
+        args.group_size is None
+        and args.num_group is None
+        and constants.DEFAULT_NUMGRP <= students_list_length / 2
+    ):
```
Lancasterwu: In parse_argument.py if you set `required=True`, you don't need to check `args.group_size is None` and `args.num_group is None`here.

gatorgrouper_cli.py
```diff
@@ -48,40 +49,65 @@
...
+        elif (
+            GG_ARGUMENTS.grouping_method == constants.ALGORITHM_ROUND_ROBIN
+            and GG_ARGUMENTS.group_size is None
+            and GG_ARGUMENTS.num_group is None
+        ):
+            GROUPED_STUDENT_IDENTIFIERS = group_rrobin.group_rrobin_num_group(
+                SHUFFLED_STUDENT_IDENTIFIERS, constants.DEFAULT_NUMGRP
+            )
+        # If the user wanted to sort using the graph algorithm and provided the
+        # number of groups
```
Lancasterwu: In parse_argument.py if you set required=True, you don't need to check `args.group_size is None` and `args.num_group is None`here.

gatorgrouper/utils/parse_arguments.py
paladp: Default values were removed as a result of the changes required in the logic of the if statements in gatorgrouper_cli. If we can find a way to have defaults in, but also check for explicitly set values that would be ideal, but I don't know of a good non-intrusive way to do that.

gatorgrouper/utils/parse_arguments.py
paladp: For both of these comments, setting both of these arguments as a required will mess with the logic. Currently we check for which value (if any) was set explicitly by the user. If they set both, which one do we pick? How do they communicate that to us? What if the two values they pick are incompatible? These are questions that would come up if we set them to required. Ultimately the fact that these options can take the value of `None` is the centerpiece of the logic. I do think that it's ultimately not the best thing to do, but am unsure of what other way we should do it. If you have a plan or general design idea, we could talk about it and maybe prototype it in another branch?

Lancasterwu: I'll explain that later when we meet.

paladp:
Need a rereview because I reverted certain files to their master versions in order to merge this PR and start work on a more effective change to the gatorgrouper_cli and parse_argument files.

@aubreypc @huangs1 @Lancasterwu

aubreypc:
@Michionlion This is ready to merge!

PR #254, opened 2019-03-12
Adding tables within database and changes to .gitignore

wattob:
# Description of Pull Request

Added the migrations for the db.sqlite3 and changed the .gitignore to maintain the tables.

## Type of Change

- [ ] Bug fix
- [ ] Breaking change
- [ ] New feature
- [ ] Documentation update
- [ ] Other


.gitignore
```diff
@@ -20,6 +20,7 @@ secret_key.py
 __pycache__/
 local_settings.py
 media
+db.sqlite3
```
Michionlion: db.sqlite3 should not be ignored according to @sutterj.

sutterj: @Michionlion 
In order to prevent people from committing testing work to the database, we have committed it in its initialized form and are now adding it to the ignore.  If anyone changes the schema, they will have to force add changes, but we don’t want emails and stuff used in testing to end up in the database (which is what is happening now).

Michionlion: Once a file is tracked, even if it is later ignored it will still be included by default (and does not need to be force added). So this will not have the desired effect.

sutterj: @Michionlion 
The file needs to be in the repo, but we want it to be in a state where it is not going to get changed when someone does a -A.  What would you suggest to accomplish this?

Michionlion: Don't use `-A`. :smiley:

sutterj: I would like for that to be the convention that everyone would follow, but I don’t have much faith in the complete adoption of this.  Would untracking the file, updating the ignore, and then forcing the file initially work or is that still the same as the current situation?

Michionlion: No, that would result in the same situation. As long as a file is tracked, updates to that file will be added if they match the path given to `git add` (`-A`, `*`, or anything else).

sutterj: Alright.  @wattob take the db.sqlite3 out of the ignore and @Michionlion we’ll just have to make sure not to approve any pull requests that include changes to the db.  We should make merging this pull request a priority so that way everyone has the correct and empty db.

Lancasterwu:
-Missclicked

Lancasterwu:
-oops, missclicked

.gitignore
wattob: @Michionlion @sutterj I have made the change you suggested and removed `db.sqlite3` from the .gitignore. Please let me know if any other changes are needed.

PR #259, opened 2019-03-13
Updated Django MVP

### LINKED ISSUE #161: Write test cases for Django models and views

barrezuetai:
# Description of Pull Request
This pull request brings new additions and fixes to issues we've had. Here are a list of major and minor changes that have been implemented in this issue.

## New Registration/Login Method
This pull request adds a simple way for users to log in. There's a register page that requires first name, last name, and email in order to create a Professor user. Once that user is created, they can log in and start accessing the pages that are used to create classes, assignments, and add students. 

## Profile Page for user
The way it is now, the profile page is a very basic page that displays the professor's current classes and assignments. The page also provides the professor with a list of students. (We can add more to this after it's been merged so that it looks prettier).

## Grouping Students
This branch integrates gatorgrouper round robin method of grouping and provides the user with a page that allows them to easily create groups for their assignments. It lets them display it, and if they want, they can save that group. It's been implemented in a way where one assignment can only have a groups created for it once. The user can choose the group size by entering a number or clocking on the arrows in the number field.

## Group Persistence
Once the user clicks the save button in the create groups page, they can access the groups that they've save in the view groups page which can be accessed by clicking on an assignment name in their profile.

##Refactored Templates
All of the templates that use forms were refactored to use Django's template code and formmodels. This resulted in much cleaner looking html. (These sites can still be styled further to look prettier).

## Only authorized user access
After this pull request, only authorized users will be allowed to access the forms and profile pages.

## More Test Cases for Django
More test cases have been added for models and views.

## Drop Down Menu
A drop down menu has been added to the nav bar for easier access to other pages.

## Smaller Changes
There were a few character symbols that kept appearing in each page. This was a result of extra html being found in the base. The extra characters have been removed. All instances of absolute references have been switched to referencing via Django's template code. Filled in all of the links that didn't take the user anywhere. Changed the font color in the home page so that the name gator grouper could be better read.

Fixes #161 

## Type of Change

- [x] Bug fix
- [x] Breaking change
- [x] New feature
- [ ] Documentation update
- [ ] Other


.pylintrc
```diff
@@ -7,7 +7,7 @@ extension-pkg-whitelist=
 
 # Add files or directories to the blacklist. They should be base names, not
 # paths.
-ignore=CVS
+ignore=CVS, 0001_initial.py, test_group_method.py, group_graph.py, 0002_auto_20190313_1424.py
```
sutterj: Is it supposed to be CVS or should it be CSV?

barrezuetai: CVS, that's how it came.

barrezuetai:
@sutterj So I should delete the db and migrations from this pull request is what you're saying?

sutterj:
@barrezuetai Just delete the db.  The migrations won’t hurt.

sutterj:
@barrezuetai The reason we are deleting the db is so we don’t commit all of the emails and names that were used when testing.  All of the users that have logged into the page are visible if you browse the db and we don’t want to put that out there.

barrezuetai:
@sutterj Gotcha, Just deleted, committed and pushed

barrezuetai:
> Need this in the master

We need more approvals. Can you approve?

PR #268, opened 2019-03-14
Web/feature/oauth

### LINKED ISSUE #145: Allow users to log in with Google / GitHub
In order to simplify the login process, we should use OAuth to allow users to sign in with their existing Google or GitHub accounts.

This blog post might help: https://medium.com/trabe/oauth-authentication-in-django-with-social-auth-c67a002479c1

Additionally, we could allow professors to configure their courses to add a college domain name, so that users who have an email with that domain attached to their Google account are allowed to join the course.

finneyj2:
# Description of Pull Request

This pull request has a working version of the oauthentication.
Fixes #145

- [ ] Bug fix
- [ ] Breaking change
- [X] New feature
- [ ] Documentation update
- [ ] Other


gator_grouper/settings.py
```diff
@@ -48,6 +44,37 @@ def find_or_create_secret_key():
     return key
 
 
+def find_oauth_secret_key():
```
Michionlion: I think you need to add some way to set this from an environment variable as well, so that it can be specified for Travis, as that is currently why it is failing.

gator_grouper/settings.py
```diff
@@ -14,9 +14,34 @@
...
+    Otherwise, generate a new secret key, save it in secret_key.py, and return the key.
+    """
+    SECRET_KEY_DIR = os.path.dirname(__file__)
+    SECRET_KEY_FILEPATH = os.path.join(SECRET_KEY_DIR, "secret_key.py")
+    sys.path.insert(1, SECRET_KEY_DIR)
+    if os.path.isfile(SECRET_KEY_FILEPATH):
+        # pylint: disable=import-error
+        from secret_key import OAUTH_SECRET_KEY as key
+
+        return key
```
Michionlion: You need to have a return (or exception) case when the oauth key cannot be found.

szklenskij: Travis does not like the `raise Exception("Couldn't find the oauth key")` as per what @finneyj2  has informed me

finneyj2: @Michionlion our subsequent attempts to add an exception case have only led to the pytest cov. error returning

PR #269, opened 2019-03-14
Command line interface method

### LINKED ISSUE #255: Provide Interface of All Group Algorithm for CLI
I would suggest adding a single interface method to this module that takes two string constants from `constants.py` (one for grouping method and the other for whether to use the given number for group size or group number) and then calls the correct grouping function based on those arguments. This will make all API and testing calls the same function, enabling better, easier to read code.

_Originally posted by @Michionlion in https://github.com/GatorEducator/gatorgrouper/pull/222_

Lancasterwu:
# Description of Pull Request

Fixes #255 

Create a new `run_argument` method and test case for `run.py` as refactoring the structure of the program so those conditional logic can be tested.
Also fixed a bug in parse_argument and constant that default value still exist when `group-size` is removed and `num-group` is required.

## Type of Change

- [x] Bug fix
- [ ] Breaking change
- [ ] New feature
- [x] Documentation update
- [x] Other (refactor)

## Tags

@Lancasterwu


Lancasterwu:
@Michionlion Hi I think this PR is ready to be merged.

PR #271, opened 2019-03-15
Fixed model access bug/ updated database with new changes

barrezuetai:
# Description of Pull Request
@griffinh1 and Alex K found a bug yesterday where users can access other users' class, assignment, and student information through form options. This bug fixes that issue by adding new custom forms that filter objects by the current user that's logged in.

In order to do this, there needed to be changes to the database schema which is why this update includes a new updated/cleaned database and migrations folder.

## Type of Change

- [x] Bug fix
- [ ] Breaking change
- [ ] New feature
- [ ] Documentation update
- [ ] Other


rauniyars:
Changes look good

sutterj:
@barrezuetai Just update the branch with master and then this is good to merge.

gatorgrouper/utils/group_creation.py
```diff
@@ -63,7 +63,7 @@ def group_rrobin_num_group(responses, numgrps):
     indices = list(range(0, numgrps))
     random.shuffle(indices)
     target_group = itertools.cycle(indices)
-    priorityColumn = random.randint(1, len(responses[0]) - 1)
+    priorityColumn = random.randint(0, len(responses[0]) - 1)
```
aubreypc: Correct me if I'm wrong, but wouldn't changing this to 0 mean that the first column (i.e. the names column) could be chosen as priority?

barrezuetai: If it's kept at 1, the grouping won't work for the web app. It sends us to an error page. Right now there's now survey, so there's no way to get traits associated with students. For grouping students through the web app, the names are currently the priority, since there are no other columns.

aubreypc: It might be best to instead return a random grouping when there is no data besides the names, either via a check on the length of responses[0] or exception handling for the IndexError. My guess is that since `"studentname" is not True` on line 72 this just acts as a random grouping anyway in that case, but if the `is True` check is ever changed to an `== True` or other kind of check changing the random range could lead to unexpected behavior.

barrezuetai: We can push this and then create a new pull request that handles that change?

PR #272, opened 2019-03-15
Deploying web app

alexheinle:
# Description of Pull Request

The deploying team has followed the steps to deploy the GatorGrouper Web Application.  All steps were followed in order to have automatic deploys.  We created a section in the ReadMe that includes directions and steps to follow incase the automatic deploy fails.

The site is not fully deployed because Professor Kapfhammer had to create an account on AWS. He is still waiting for the confirmation to deploy.

## Type of Change

- [ ] New feature


README.md
```diff
@@ -349,6 +349,201 @@ foreign keys using `django.db.models.ForeignKey`. The different entities within
 the classes are linked by a means one-to-many, many-to-one, and many-to-many
 relationship.
 
+### Set Up a Python Virtual Environment with Django
```
aubreypc: Since our README already has a section on setting up pipenv, these virtualenv instructions might be redundant

sutterj: Since you already have pipenv you can just use the command `pipenv lock -r > requirements.txt` to generate the requirements.txt.  Elastic beanstalk does not natively support pipenv so you need the requirements file.

corlettim: While our README does have instructions for setting up pipenv, the pipenv instructions are for installing rather than creating and putting your terminal in a virtual environment. That is why we figured to include this with the other instructions for deployment. I also think that it would be helpful to repeat (if it truly is), that way a user does not have to scroll up and down to find the virtual environment instructions and back to the deployment ones.

Michionlion: This should use pipenv to create the virtualenv.

PR #278, opened 2019-03-15
Generalize interface for Both CLI and WI

Lancasterwu:
# Description of Pull Request

* Fix a bug caused by the [previous PR](https://github.com/GatorEducator/gatorgrouper/pull/271#discussion_r266013148), which broken `rrobin` method since the priority column should not include students' name. 
* Provide the exception handling when `rrobin` is called with the list of list that only contains student names. This fixes the original reason for making the bug above. If the input list only contains names, it will run `random`.
* Provide the generalized run interface to improve the structure in CLI and make all functions can be called easier in web interface.
* Provide test case for the interface method above.
* Fixed a bug that `random` method will always provide the same output if it is used by web interface

## Type of Change

- [x] Bug fix
- [ ] Breaking change
- [x] New feature
- [ ] Documentation update
- [x] Test case

## Tags

@Lancasterwu @paladp @barrezuetai


gatorgrouper/utils/run.py
```diff
@@ -25,23 +25,51 @@ def run_arguments(GG_ARGUMENTS, preference=None):
...
+    preferences=None,
+    preferences_weight=None,
+    preferences_weight_match=None,
+    objective_weights=None,
+    objective_measures=None,
+):
+    """ Run conditional logic statment to ran different methods """
+    students_reshuffle = group_creation.shuffle_students(students)
+    if method == constants.ALGORITHM_ROUND_ROBIN:
+        if (len(students_reshuffle[0])) < 2:  # Note that this only checks the first student
```
Michionlion: Unneeded extra parenthesis around the `len()` call here.

gatorgrouper/views.py
```diff
@@ -280,7 +280,7 @@ def create_groups(request):  # pylint: disable=too-many-locals
             student_list = []
             for name, obj in student_list_dict.items():
                 student_list.append(name)
-            groups = group_rrobin_num_group(student_list, num_of_groups)
+            groups = input_interface(student_list, "rrobin", num_of_groups)
```
Michionlion: You should use the `constants.py` algorithm value here, that is why it exists!

tests/test_run.py
```diff
@@ -41,3 +41,10 @@ def test_run_rrobin(generate_csv):
     input_arguments = parse_arguments.parse_arguments(command)
     output = run.run_arguments(input_arguments)
     assert len(output) == 3
+
+
+def test_run_rrobin_to_random():
+    """ Test if the main program can run rrobin then transfer to random """
+    namelist = [["A"], ["B"], ["C"], ["D"]]
+    output = run.input_interface(namelist, "rrobin", 2)
```
Michionlion: Does this test ensure that `random` is actually run?

Lancasterwu: Yea it is running `random` because if we run `rrobin` with list like `[["A"], ["B"], ["C"], ["D"]]` we will get index out of bound error.

gatorgrouper/utils/run.py
```diff
@@ -52,7 +52,9 @@ def input_interface(
     """ Run conditional logic statment to ran different methods """
     students_reshuffle = group_creation.shuffle_students(students)
     if method == constants.ALGORITHM_ROUND_ROBIN:
-        if (len(students_reshuffle[0])) < 2:  # Note that this only checks the first student
+        if (
+            len(students_reshuffle[0]) < 2
+        ):  # Note that this only checks the first student
```
Michionlion: Is this from black? it should look like `if len(students_resuffle[0]) < 2:`

gatorgrouper/utils/run.py
```diff
@@ -9,6 +9,7 @@
 from gatorgrouper.utils import display
 from gatorgrouper.utils import constants
 
+# pylint: disable=bad-continuation
```
Michionlion: This should be positioned where it is needed -- here it disables this check for the entire file!

gatorgrouper/utils/run.py
Lancasterwu: Yea it is from black, I changed it and black reformated it back

gatorgrouper/utils/run.py
Lancasterwu: The reason I put it here is due to that `pylint` warms bad-continuation from line 45 to 52. I don't want to put `pylint disable` for 8 lines.

gatorgrouper/utils/run.py
Michionlion: I just checked and black should not reformat that code. The below code, matching exactly the part that matters, is not reformatted.

```
students_reshuffle = [[23], ["t"]]
if len(students_reshuffle[0]) < 2:
    print("tested")

```

Michionlion: Ah, it's because of the tailing comment. That should be on the line above the `if`, as should the `pylint: disable` statements scattered throughout this code.

Lancasterwu: I have tried again, black does change it. Is it possible due to the version? I have `black, version 19.3b0`

Lancasterwu: oh, thanks. I'm going to change it.

Lancasterwu: fixed

gatorgrouper/utils/run.py
Michionlion: Then put it above the method that those lines are in, that way it is only that method where it is ignored.

Lancasterwu: resolved

Michionlion: @Lancasterwu can you also quickly remove those trailing `pylint: disable`s and move them above their lines? Using trailing comments does not make for clean code.

Lancasterwu: Sure

Lancasterwu: fixed

PR #279, opened 2019-03-16
AWS deploy

### LINKED ISSUE #245: Create a plan for deploying the web application
# Description of the Issue

In order for our web application to be publicly available, we will need to figure out a way to deploy it.

- Choose a hosting provider
- [Configure Travis for deployment](https://docs.travis-ci.com/user/deployment/)
- Deploy our web app

sutterj:
# Description of Pull Request

Fixes code related to an inability to deploy correctly.

Fixes #245 

## Type of Change

- [x] Bug fix
- [ ] Breaking change
- [x] New feature
- [ ] Documentation update
- [ ] Other 

## Tags

@gkapfham @Michionlion @aubreypc @barrezuetai


barrezuetai:
Is it alright that the build is failing?

sutterj:
I am fixing stuff that is related to what I changed, but there are other tests that are failing that are already a part of master.

Michionlion:
The build is failing because the characters in the secret key for travis are not escaped correctly.

sutterj:
I fixed that as part of this PR.

sutterj:
The problem that I am now seeing is with test_forms.py.

Michionlion:
Ah, I just looked and that just seems to be simple black and pylint stuff.

Michionlion:
With that merge from master, @sutterj, migrations will likely need to be regenerated.

gator_grouper/settings.py
```diff
@@ -168,3 +126,22 @@ def find_or_create_secret_key():
 LOGIN_REDIRECT_URL = "Gatorgrouper-home"
 
 LOGIN_URL = "login"
```
Michionlion: `LOGIN_URL` is given twice, as is `LOGIN_REDIRECT_URL` (thrice actually)

sutterj: I didn't manually edit any of the travis file.  The only thing I did was run `travis encrypt SOMEVAR="secretvalue"`.  It must have changed the whole file when I did that.  I can change it back.

sutterj: As for the urls, those were all added by someone else.  I'll get rid of the duplicates.

.travis.yml
```diff
@@ -17,6 +17,10 @@ env:
     - LANG=en_US.UTF-8
     - LC_ALL=en_US.UTF-8
     - secure: 2c0250fb-dfb1-45c9-abaf-cd3d874326c5
+    - secure: 2c0250fb-dfb1-45c9-abaf-cd3d874326c5
```
Michionlion: Should these lines be repeated? I'm not sure how the secured version of travis works, but this doesn't seem right...

Michionlion:
@sutterj, it seems that the problem below may be created from the fact that there was no initial `password` field in the given form, although I'm not positive.

```
self = <tests.test_forms.TestCustomUserChangeForm object at 0x7f526e212160>
    def test_valid_data(self):
        """undocumented"""
        form = CustomUserChangeForm(
            {
                "email": "testuserl@test.com",
                "first_name": "Spencer",
                "last_name": "Huang",
                "password1": "testpassword1",
                "password2": "testpassword1",
            }
        )
>       assert form.is_valid() is True
tests/test_forms.py:48: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
.venv/lib/python3.7/site-packages/django/forms/forms.py:185: in is_valid
    return self.is_bound and not self.errors
.venv/lib/python3.7/site-packages/django/forms/forms.py:180: in errors
    self.full_clean()
.venv/lib/python3.7/site-packages/django/forms/forms.py:381: in full_clean
    self._clean_fields()
.venv/lib/python3.7/site-packages/django/forms/forms.py:402: in _clean_fields
    value = getattr(self, 'clean_%s' % name)()
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
self = <CustomUserChangeForm bound=True, valid=True, fields=(email;first_name;last_name;password)>
    def clean_password(self):
        # Regardless of what the user provides, return the initial value.
        # This is done here, rather than on the field, because the
        # field does not have access to the initial value
>       return self.initial["password"]
E       KeyError: 'password'
.venv/lib/python3.7/site-packages/django/contrib/auth/forms.py:152: KeyError
```

sutterj:
That is what it seemed like to me as well.  There is just email, first, and last in the `form.py` file and not password, but I wasn't sure if the part where it called the user model somehow deals with the password.

Michionlion:
I think that error is perhaps best solved by one of the people who originally developed that test case. I think @enpuyou would know who that was. In the mean time, I think that if that test is skipped and the rest of the tests pass, we should be good to merge to master.

sutterj:
@Michionlion Should we ignore limiting for the migration files since they are auto-generated?

Michionlion:
> @Michionlion Should we ignore pylint for the migration files since they are auto-generated? That is what is currently causing the fail and running black is not enough to fix it.

I've been doing some research and it appears, due to a bug with `pylint`, there is no way to ignore directories. I'm trying to come up with a regex that will only match django migration basenames, do migrations always start with four digits?

sutterj:
@Michionlion As far as I know the ones that are autogenerated always start with 4 digits and they can be renamed to whatever you want them to as long as the same number does not appear twice.

enpuyou:
>@sutterj, it seems that the problem below may be created from the fact that there was no initial password field in the given form, although I'm not positive.

The valid form seems to be <CustomUserChangeForm bound=True, valid=True, fields=(email;first_name;last_name;password)>

I think you are right about that. This test case was first created attempting to test when this form is used in the `views.py` where registering account requires two password to be put in. I will work on fixing this now and it should be a easy fix. 

One quick question, when I try to run pytest, it returns this error:
```   SECRET_KEY = os.environ["GATOR_GROUPER_KEY"]
  File "/Users/enpuyou/.local/share/virtualenvs/gatorgrouper-p6ulwK35/lib/python3.6/os.py", line 669, in __getitem__
    raise KeyError(key) from None
KeyError: 'GATOR_GROUPER_KEY' ```
Am I supposed to run any command in terminal to solve this?

sutterj:
@enpuyou In order to run the tool locally you have to create a folder in the root of the repo called .env and store both the GATOR_GROUPER_KEY and GATOR_GROUPER_OAUTH keys that were posted in slack.

Michionlion:
@gkapfham @sutterj, this is not the place to note this, but I've discovered that we cannot fix the current security alert for `pyyaml` because one of the packages depends on `pyyaml <=4.13` and `>=4.10`

Michionlion:
@sutterj I'm pushing this to master now since it passes travis while master may not.

Tekulearn:
SOCIAL_AUTH_GOOGLE_OAUTH2_SECRET 
how to get this key?

Michionlion:
@Tekulearn Unfortunately, this project is unmaintained at the moment. The key you mentioned is, I believe, an authorization key you would need to obtain from Google, likely through their Developer Console.

Tekulearn:
Can we run the program with out using Google Authenticator code?

Michionlion:
You may be able to set it to `""`, but there's no guarantees. You're largely on your own in understanding and running this project, and I'd actually suggest forking it and looking into the source code for more information. There are a lot of features that may not be necessary for your use case, and you could remove those (like the Google Authentication, which I believe is for pulling a student file from a Google Sheets document).

PR #281, opened 2019-03-16
QuickFix/skipped valid form test in `test_forms.py`

### LINKED ISSUE #280: Fix skipped valid form test in `test_form.py`
## Fix skipped test

There is currently a skipped test on master that must be fixed. Below is some correspondence that occurred on PR #279 about this issue. It is related to `test_forms.py`.

### Assigned Developers
@enpuyou

enpuyou:
# Description of Pull Request

Fixes #280 

## Type of Change

- [x] Bug fix
- [ ] Breaking change
- [ ] New feature
- [ ] Documentation update
- [ ] Other 

## Tags

@enpuyou


tests/test_forms.py
```diff
@@ -42,8 +41,6 @@ def test_valid_data(self):
                 "email": "testuserl@test.com",
                 "first_name": "Spencer",
                 "last_name": "Huang",
```
Michionlion: It's likely that you will still need a password entry here...

enpuyou: I think they work the same. This function is not called in any files now and in `forms.py` the field only requires email, first_name, and last_name. But I added the password entry in the most recent commit, and it passes as well.

Michionlion: @enpuyou the newest commit doesn't pass, it fails with the same error... Perhaps you should coordinate with whoever put together the forms? Would that have been @barrezuetai or @aubreypc?

enpuyou: Yes, I just saw that. It's weird that the test case passed in my local. I think this form is created by @barrezuetai. And I think I know why this is happening. Let me try something else see if it works...

tests/test_forms.py
```diff
@@ -41,13 +41,25 @@ def test_valid_data(self):
...
+        form = CustomUserChangeForm(
+            {
+                'email': 'normaluser@user.com',
+                'first_name': 'testfirst',
+                'last_name': 'testlast',
+                # 'password': 'testpassword',
+            }
+            , instance=user
+        )
+>>>>>>> 8fe81fdaf00d27045757d66368fb90aa9ee936e7
```
sutterj: @enpuyou You’ll have to go through and manually remove the merge markers.

enpuyou: Just removed it! Forgot to save files when resolving the merge conflicts. The newest commit should pass.

sutterj:
@enpuyou Is there a reason why it got changed from entering two passwords to one?  Does the form also check to see that they match or was it like initially in error?

PR #291, opened 2019-03-26
Fix arguments passed to graph algorithm

aubreypc:
# Description of Pull Request

Allows the graph algorithm to be run from the command line with objective measures, weights, and preferences handled correctly 

## Type of Change

- [ ] Bug fix
- [ ] Breaking change
- [ ] New feature
- [ ] Documentation update
- [ ] Other


quigley-c:
If I'm not mistaken the fix just matches the types passed as arguments, correct?

aubreypc:
@quigley-c Yeah, it just ensures that the lists for weights and measures can be specified on the command line, and sets up the preferences dictionary with the right format

gatorgrouper_cli.py
```diff
@@ -22,7 +22,13 @@
...
         preference = None
     else:
-        preference = dict(read_student_file.read_csv_data(GG_ARGUMENTS.preferences))
+        preference = read_student_file.read_csv_data(GG_ARGUMENTS.preferences)
+        preference = [
+            [entry for entry in row if entry != ""]
+            for row in preference
+            if len(row) > 1
+        ]
+        preference = {row[0]: set(row[1:]) for row in preference}
```
Michionlion: You should use a temporary variable for the intermediate decoding steps here; using `preference` could cause problems in the future, and confuses what the final value should be.

aubreypc: Rather than use a temp variable I changed the `preferences` within the list comprehension to call `read_csv_data` directly. Does that help readability (or make it worse)?

Michionlion: It actually makes it worse -- although it may make the code longer or less performant, two different names like `csv_data` or `validated_data` will make it much clearer. `preference` should only be set once.

aubreypc: The latest commit distinguishes between the list `preference_list` and dict `preference`, setting `preference` only once -- is that in line with your suggestion or am I misunderstanding?

Michionlion: Yes, perfect!

